PR #19, opened 2020-07-13
[io] Make io traits consistent with std::io

mcy:
See #20.


src/io.rs
```diff
@@ -341,6 +341,106 @@ impl Write for &'_ mut [u8] {
...
+    }
+
+    /// Consumes `n` bytes from the underlying buffer.
+    ///
+    /// If `n` bytes are unavailable, `None` is returned.
+    pub fn consume(&mut self, n: usize) -> Option<&mut [u8]> {
+        if self.buf.len() - self.cursor < n {
+            return None;
+        }
+        let end = self.cursor + n;
```
gkelly, line 395: How about:

```rust
let end = self.cursor + n;
if end > self.buf.len() {
  return None;
}
```
Just reads a little cleaner and uses the `end` value instead of constructing the constraint independently.

mcy:
Making the repo public broke this PR's link to my fork. #20 replaces it.

src/io.rs
mcy: This is fixed in #20.

PR #20, opened 2020-07-15
[io] Make io traits consistent with std::io

mcy:
Changes include:
- `&mut impl Read: Read` and `&mut impl Write: Write`.
- Everything that can take `Read`/`Write` takes it by value.
- `Read` and `Write` are object-safe.
- Addition of `io::Cursor`, which simplifies using `Write` on buffers.
- Use of `io::Cursor` everywhere where we used to do ad-hoc buffer I/O.


src/io.rs
```diff
@@ -225,11 +235,28 @@ pub trait Read<'a> {
     fn remaining_data(&self) -> usize;
 
     /// Reads a little-endian integer.
-    fn read_le<I: LeInt>(&mut self) -> Result<I, Error> {
+    fn read_le<I: LeInt>(&mut self) -> Result<I, Error>
+    where
+        Self: Sized,
```
lenary, line 240: Q: Why is this constraint required here and in `Write::write_le`?

src/io.rs
```diff
@@ -225,11 +235,28 @@ pub trait Read<'a> {
...
+impl<'a, R: Read<'a> + ?Sized> Read<'a> for &'_ mut R {
+    #[inline]
+    fn read_bytes(&mut self, n: usize) -> Result<&'a [u8], Error> {
+        R::read_bytes(*self, n)
+    }
+
+    #[inline]
+    fn remaining_data(&self) -> usize {
+        R::remaining_data(*self)
+    }
```
lenary, line 257: Don't you want an implementation of `read_le` which uses `R::read_le` rather than using `Self::read_bytes`?

src/io.rs
```diff
@@ -274,11 +301,28 @@ pub trait Write {
...
+impl<W: Write + ?Sized> Write for &'_ mut W {
+    #[inline]
+    fn write_bytes(&mut self, buf: &[u8]) -> Result<(), Error> {
+        W::write_bytes(*self, buf)
+    }
+
+    #[inline]
+    fn remaining_space(&self) -> usize {
+        W::remaining_space(*self)
+    }
```
lenary, line 323: Again, don't you want to pass `write_le` down to `W::write_le`?

src/io.rs
```diff
@@ -341,6 +341,106 @@ impl Write for &'_ mut [u8] {
...
+///
+/// let msg_bytes = cursor.take_consumed_bytes();
+/// assert_ne!(msg_bytes.len(), 0);
+/// ```
+///
+/// [`Write`]: trait.Write.html
+/// [`ToWire`]: ../protocol/wire/trait.ToWire.html
+pub struct Cursor<'a> {
+    buf: &'a mut [u8],
+    // Invariant: cursor <= buf.len().
```
lenary, line 378: Nice

src/io.rs
mcy: This is so that the traits are object-safe (i.e., `dyn Read<'a>` and `dyn Write` are well-formed types). Generic functions are disallowed on object-safe traits (unsurprising, because it would require runtime monomorphization). The `Self: Sized` bound removes this requirement.

Mind, because `&mut impl Read: Read`, then `&mut dyn Read` can be passed into any function taking `impl Read`, and, because `&mut dyn Read` (unlike `dyn Read`) is `Sized`, you can call `read_le()` on it. I should probably mark the default impl of `read_le` as inline, too, since it's a trivial function and it would cut down on binary bloat...

src/io.rs
mcy: Inlining should make this a non-issue, since all those functions are basically trivial. If we see a problem with code bloat, we have ways to deal with it.

lenary: Yeah, this isn't about inlining, this is about method resolution. :+1:

src/io.rs
mcy: See above.

PR #22, opened 2020-07-15
[test] Factor out hardware fakes for re-use

mcy:
Signed-off-by: Miguel Young de la Sota <mcyoung@google.com>

### DIFF
--- a/src/hardware.rs
+++ b/src/hardware.rs
@@ -122,3 +122,63 @@ impl Flash for [u8] {
         Ok(())
     }
 }
+
+#[cfg(test)]
+pub(crate) mod fake {
+    use core::convert::TryInto;
+    use core::time::Duration;
+
+    /// A fake `Identity` that returns fixed values.
+    pub struct Identity {
+        firmware_version: Vec<u8>,
+        unique_id: Vec<u8>,
+    }
+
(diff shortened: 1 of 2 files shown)


src/hardware.rs
```diff
@@ -122,3 +122,61 @@ impl Flash for [u8] {
...
+
+    impl Identity {
+        /// Creates a new `fake::Identity`.
+        pub fn new(firmware_version: &[u8], unique_id: &[u8]) -> Self {
+            let mut firmware_version = firmware_version.to_vec();
+            while firmware_version.len() < 32 {
+                firmware_version.push(0);
+            }
+            Self {
+                firmware_version,
```
lenary, line 145: you might want to slice this to ensure it's no longer than 32 bytes when stored, which might simplify returning it?

mcy: The implementation of Identity below slices it, but I'll go ahead and truncate it for good measure.

I'd like to have a Box<[u8; 32]> but there isn't a good way to get there from a Box<[u8]> right now.

lenary: oh i see, right, so this doesn't actually help much with the `firmware_version` function implementation.

PR #28, opened 2020-07-16
[ring] Make crypto::ring part of Manticore's API

mcy:
This PR also moves around some Cargo features. In particular, `std` and `ring` are both included by default, because they are necessary for tests (Cargo does not allow specifying additional test-only features, and this behavior seems consistent with other crates).


lenary:
I'm not sure I understand the implications of this enough to review it.

PR #29, opened 2020-07-16
Remove unused io::Write::remaining_space.

osenft:


src/io.rs
```diff
@@ -303,9 +303,6 @@ pub trait Write {
     /// until completion or return an error.
     fn write_bytes(&mut self, buf: &[u8]) -> Result<(), Error>;
 
-    /// Returns the number of bytes still available to read.
```
mcy, line 306: You can tell I didn't even think about it when I added this because this merely says "bytes available to read"!

PR #31, opened 2020-07-16
[io] Add std::io support (as much as we can, anyway)

mcy:


src/io.rs
```diff
@@ -341,6 +361,38 @@ impl Write for &'_ mut [u8] {
...
+///
+/// [`std::io::Write`]: https://doc.rust-lang.org/std/io/trait.Write.html
+/// [`manticore::io::Write`]: trait.Write.html
+/// [`write_bytes()`]: trait.Write.html#tymethod.write_bytes
+/// [`write()`]: https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write
+#[cfg(feature = "std")]
+pub struct StdWrite<W>(pub W);
+
+#[cfg(feature = "std")]
+impl<W: std::io::Write> Write for StdWrite<W> {
```
osenft, line 382: Could this not be something like this:
```
impl Write for T where T : std::io::Write {
```

src/io.rs
```diff
@@ -341,6 +361,38 @@ impl Write for &'_ mut [u8] {
...
+    fn write_bytes(&mut self, mut buf: &[u8]) -> Result<(), Error> {
+        use std::io::ErrorKind;
+        loop {
+            match self.0.write(buf).map_err(|e| e.kind()) {
```
osenft, line 386: Do we have to implement an explicit break from the loop when buf is exhausted?

src/io.rs
```diff
@@ -130,6 +130,9 @@ pub enum Error {
     /// This is typically a fatal error, since it is probably not possible
     /// to re-allocate that unrelying buffer.
```
alphan, line 131: typo: unrelying

src/io.rs
```diff
@@ -130,6 +130,9 @@ pub enum Error {
...
     BufferExhausted,
+
+    /// Indicates that an unspecified, internal failure occured.
```
alphan, line 134: typo: occured

src/io.rs
```diff
@@ -218,10 +221,19 @@ impl LeInt for u64 {
 
 /// Represents a place that bytes can be read from, such as a `&[u8]`.
 ///
-/// Unlike [`std::io::Read`], this trait is intended for performing exact reads
-/// out of a buffer with a fixed lifetime; hence, the additional lifetime
-/// argument. In other words, a `manticore::io::Read` owns all of the data it
-/// returns.
+/// Types which implement this trait enable *zero copy reads*, that is, there
+/// a read opeartion does not need to allocate memory to perform a read, since
```
alphan, line 225: typo: opeartion

src/io.rs
```diff
@@ -218,10 +221,19 @@ impl LeInt for u64 {
 
 /// Represents a place that bytes can be read from, such as a `&[u8]`.
 ///
-/// Unlike [`std::io::Read`], this trait is intended for performing exact reads
-/// out of a buffer with a fixed lifetime; hence, the additional lifetime
-/// argument. In other words, a `manticore::io::Read` owns all of the data it
-/// returns.
+/// Types which implement this trait enable *zero copy reads*, that is, there
```
alphan, line 224: ~there~?

src/io.rs
alphan: +1?

src/io.rs
osenft: Or maybe that?

```
impl<T: std::io::Write> Write for T {
```

src/io.rs
```diff
@@ -341,6 +361,38 @@ impl Write for &'_ mut [u8] {
...
+/// This type is provided instead of implementing [`manticore::io::Write`]
+/// directly for every [`std::io::Write`] due to trait coherence issues
+/// involving the blanket impl on `&mut _`.
+///
+/// [`std::io::Write`]: https://doc.rust-lang.org/std/io/trait.Write.html
+/// [`manticore::io::Write`]: trait.Write.html
+/// [`write_bytes()`]: trait.Write.html#tymethod.write_bytes
+/// [`write()`]: https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write
+#[cfg(feature = "std")]
+pub struct StdWrite<W>(pub W);
```
osenft, line 379: I think this need to be `pub &'a mut W`, as otherwise we'd move `W` into `StdWrite`, which means that we couldn't use it anymore. Or am I missing something?

osenft: I ended up with the following implementation for testing:
```
pub struct StdWrite<'a, W>(pub &'a mut W);

impl<'a, W: std::io::Write> Write for StdWrite<'a, W> {
    fn write_bytes(&mut self, buf: &[u8]) -> Result<(), Error> {
      let mut idx = 0;

      while idx < buf.len() {
        match self.0.write(&buf[idx..]) {
          Err(_) => return Err(Error::Internal),
          Ok(len) => idx = idx + len,
        }
      }

      Ok(())
    }
}
```

This can now be used like that:
```
    let mut file = OpenOptions::new().write(true).truncate(true).create(true).open(&output_file).expect("failed to open output file");
    file.seek(SeekFrom::Current(mem::size_of::<u16>() as i64)).expect("failed to seek in output file");
    {
        let mut stdwrite = StdWrite(&mut file);
        header.to_wire(&mut stdwrite).expect("failed to write header");
        req.to_wire(&mut stdwrite).expect("failed to write request");
    }
    file.seek(SeekFrom::Start(0)).expect("failed to seek to beginning of output file");
    ...
```

mcy: Remember that if `W: Write`, then `&mut W: Write`, too. Therefore, if you don't want to move into the `StdWrite`, you can pass in a mutable reference into it, resulting in a `StdWrite<&'a mut W>`.

src/io.rs
mcy: Yup, good catch.

src/io.rs
osenft: Ah yes, thank you!

PR #32, opened 2020-07-22
Add string serialization/deserialization to WireEnum.

osenft:


src/protocol/wire.rs
```diff
@@ -159,6 +173,84 @@ macro_rules! wire_enum {
...
+        impl core::fmt::Display for $name {
+            fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                use $crate::protocol::wire::WireEnum;
+
+                write!(f, "{}", self.name())
+            }
+        }
+
+        impl core::str::FromStr for $name {
+            type Err = ();
```
mcy, line 204: Nit: can you please provide a specific error type rather than `()`? It should just be as simple as writing
```
#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
pub WireEnumFromStrError;
```
outside of the macro definition, and referring to it here.

osenft: Ah, good call, thanks. Done.

PR #33, opened 2020-07-22
Add CLI to serialize/deserialize messages.

osenft:


Cargo.toml
```diff
@@ -51,3 +53,18 @@ std = [
   "byteorder/std",
   "ring/std",
 ]
+
+tool = [
+  "clap",
+  "serde",
+  "serde_json",
+]
```
mcy, line 61: I really, really don't like that you have to add this feature. I can see why you had to. I may change it so that these are all default features, since on-devices uses of Manticore already have to set `default-features = false`... Sigh.

Let's leave it as-is for now.

README.md
```diff
@@ -22,6 +22,9 @@ have fuzz tests for all code that handles untrusted input. These tests, and
 instructions for executing them, can be found under the `fuzz` directory.
 Fuzz tests require nightly Rust, though the library itself builds under stable.
 
+In order to build the Manticore command line tool, run
+`cargo build --features=tool`.
```
mcy, line 26: Include `--bin manticore-tool`.

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+use manticore::protocol;
+use manticore::protocol::firmware_version;
+use manticore::protocol::wire::FromWire;
+use manticore::protocol::wire::ToWire;
+use manticore::protocol::CommandType;
+use manticore::protocol::Header;
+use serde_json;
+use std::fs::OpenOptions;
+use std::io::prelude::*;
+use std::io::BufReader;
```
mcy, line 19: Could you split up these `use` statements by crate?

src/main.rs
```diff
@@ -0,0 +1,253 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#[macro_use(value_t)]
+extern crate clap;
```
mcy, line 6: I don't believe you actually need this, you should be able to just write `use clap::value_t` these days; this is an old-style macro import.

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+
+use clap::{App, AppSettings, Arg, SubCommand};
```
mcy, line 8: Can you expand this into seprarate imports? I'm trying to avoid "nested" imports except where absolutely necessary.

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+use manticore::io::StdWrite;
+use manticore::protocol;
+use manticore::protocol::firmware_version;
+use manticore::protocol::wire::FromWire;
+use manticore::protocol::wire::ToWire;
+use manticore::protocol::CommandType;
+use manticore::protocol::Header;
+use serde_json;
+use std::fs::OpenOptions;
+use std::io::prelude::*;
```
mcy, line 18: This dumps `Read` and `Write` into scope. Consider doing `use std::io::Read as _;` to import the crate's methods, but not its name.

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+use std::io::BufReader;
+
+/// Deserializes a message in JSON format from `reader` and then serializes the message in wire
+/// format to `writer`.
+fn from_json_to_wire<'a, T, R, W>(reader: R, writer: W)
```
mcy, line 23: Nit: use `'de`, which is the usual lifetime to use with `serde::Deserialize`.

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+where
+    T: ToWire + serde::de::Deserialize<'a>,
```
mcy, line 25: Can you import `Deserialize` directly? I don't think it's ambiguoous. (I suggest keeping `serde_json::Deserializer`, though, since that name is a bit more unclear.

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    let header = Header {
+        is_request: is_request,
+        command: cmd_type,
+    };
+    header
+        .to_wire(&mut stdwrite)
+        .expect("failed to write header");
+
+    if is_request {
```
mcy, line 66: You could write this as:
```
match (is_request, command_type) {
  (true, CommandType::FirmwareVersion) => ...,
  (false, CommandType::FirmwareVersion) => ...,
  _ => ...,
}
```

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+/// Deserializes a header + message in wire format from `input_file` and then prints the message
+/// to the console.
+fn print_message(input_file: &str) {
+    read_wire_and_operate!(input_file, msg, { eprintln!("{:?}", msg) });
+}
+
+fn main() {
+    let app = App::new("Manticore Tool")
```
mcy, line 157: We can keep this like this for now, but I might come back later and clean this up to use `structopt` instead (which may or may not be a bit more readable than a huge builder...).

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+}
+
+/// Macro to deserialize wire format from an input file and then run an operation on the
+/// deserialized message.
+/// Arguments:
+/// * `input_file`: Identifier to get the input file name from.
+/// * `msg`: Identifier to hold the message to operator.
+/// * `op`: Expression to execute after deserializing `msg` from `input_file`.
+macro_rules! read_wire_and_operate {
```
mcy, line 97: The inside of a macro won't be formatter by the rust fomatter (because reasons). Would you mind wrapping all these lines to 80 characters?

However, I believe you can use [erased_serde](https://docs.rs/erased-serde/0.3.12/erased_serde/) to turn this into a function.

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+                        .takes_value(true),
+                )
+                .arg(
+                    Arg::with_name("output")
+                        .short("o")
+                        .long("output")
+                        .help("output file for message data as JSON")
+                        .required(true)
+                        .takes_value(true),
+                ),
```
mcy, line 221: Do you think that, instead of using file arguments, we could use stdin/stdout? That would make the commandline parsing simpler, and it would mean you could call it like
```
manticore-tool deserialize < blah.bin
```
instead of having `print` and `tojson` be separate code. The same with `fromjson`, since you could then write
```
manticore-tool serialize > blah.bin <<< '{json here}'
```

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+        let mut read_buf = Vec::new();
+        input.read_to_end(&mut read_buf).expect("couldn't read from file");
+
+        let mut read_buf_slice = read_buf.as_slice();
+        let header =
+            Header::from_wire(&mut read_buf_slice).expect("failed to read header");
+        if header.is_request {
+          match header.command {
+            CommandType::FirmwareVersion => {
+              let $msg : protocol::firmware_version::FirmwareVersionRequest =
```
mcy, line 110: Should be `let $msg: protocol::...`. Again, this is because cargo fmt doesn't reach into macros. =(

src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+        let mut input = OpenOptions::new().read(true).open(&$input_file).expect("failed to open input file");
+
+        let mut read_buf = Vec::new();
+        input.read_to_end(&mut read_buf).expect("couldn't read from file");
+
+        let mut read_buf_slice = read_buf.as_slice();
+        let header =
+            Header::from_wire(&mut read_buf_slice).expect("failed to read header");
+        if header.is_request {
+          match header.command {
```
mcy, line 108: Again, you can unify these two expressions statements into one.

Cargo.toml
```diff
@@ -17,6 +17,8 @@ description = "A WIP implementation of the Cerberus attestation protocol"
 arrayvec = { version = "0.5.1", default_features = false }
 bitflags = "1.2.1"
 byteorder = { version = "1.3.4", default_features = false }
+clap = { version = "2.33.1", optional = true }
+serde_json = { version = "1.0", optional = true }
```
mcy, line 21: Do you think you could separate these dependencies into another stanza?

osenft: Done.

README.md
osenft: Done

src/main.rs
osenft: Oh yes, that worked, thank you!

src/main.rs
osenft: Done. Not sure why I forgot about those. Sorry.

src/main.rs
osenft: Thanks for pointing that out. Fixed.

src/main.rs
osenft: Done.

src/main.rs
osenft: Done.

src/main.rs
osenft: Done.

src/main.rs
osenft: Ah yes, good point! I hadn't thought of this. Nice trick to know!

src/main.rs
osenft: Oh yes, I did not know that. I'll look into `erased_serde` in a separate PR.

src/main.rs
osenft: Done.

src/main.rs
osenft: Done.

src/main.rs
osenft: ACK.

Cargo.toml
osenft: ACK.

src/main.rs
osenft: I had thought of that, but I don't like the idea of sending raw binary data onto the console by default. You can already do `-i /dev/stdin` and `-o /dev/stdout`. We could look into adding `-i -` and `-o -` as aliases, but I don't think that's necessary. I'd leave it as is if that's ok.

mcy: Let's leave it like this for now, but we should re-visit this interface to make it a bit less redundant.

PR #36, opened 2020-08-05
[net] Introduce transport-layer agnosticism

mcy:
This change adds a module that abstracts over a "network connection", via the `HostPort` trait. This trait is intended to replace the current ad-hoc `manticore::protocol::Header` wire format, and to implement transport-agnosticism in Manticore.

Still pending: integration into `manticore::server::handler::Handler`. I'd like to get this design triple-checked before I go and do the work to plug it in everywhere.

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -1,0 +1,348 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Networking interfaces.
+//!
+//! This module provides generic, abstract networking interfaces for sending
+//! buffers over a "network". The term "network" is used very loosely: for
+//! our purposes, a network is an object upon which we can transact encoded
+//! buffers along with the bits necessary to encode a Cerberus header.
+//!
+//! For example, if an integration wished to recieve requests from a "host"
+//! device over a SPI line, it tie up all the necessary implementation
+//! details into a [`HostPort`] implementation.
+//!
+//! [`HostPort`]: trait.HostPort.html
+
+#![allow(missing_docs)]
+
+use static_assertions::assert_obj_safe;
+
+use crate::io;
+use crate::io::Cursor;
+use crate::io::Read;
+use crate::io::Write;
+
+pub use crate::protocol::Header;
+
+/// A networking error.
+#[derive(Copy, Clone, Debug)]
+pub enum Error {
+    /// Indicates an inderlying I/O error.
+    Io(io::Error),
+    /// Indicates an error constructing (or interpreting) an abstract Manticore
+    /// header.
+    BadHeader,
+    /// Indicates that the other end of a connection is "disconnected". This
+    /// can mean anything from a connection being explicitly terminated, to
+    /// some internal timeout expiring.
+    Disconnected,
+    /// Indicates that some operation was done out of order, such as attempting
+    /// to reference part of the request once a reply has begun.
+    OutOfOrder,
+}
+
+impl From<io::Error> for Error {
+    fn from(e: io::Error) -> Self {
+        Self::Io(e)
+    }
+}
+
+/// Represents a physical port that can be used to interact with host devices.
+///
+/// This trait provides a generic mechanism for recieving and responding to
+/// requests from host devices. A value of a type implementing this trait
+/// implements the entire network stack below the Cerberus protocol: it must
+/// know how to speak the physical layer and assemble messages out of that
+/// physical layer's packets.
+///
+/// `HostPort` uses the traits [`HostRequest`] and [`HostReply`] to describe
+/// a generic state machine, where a request is processed and replied to.
+/// ```
+/// # use manticore::net::*;
+/// fn process_request(port: &mut impl HostPort) -> Result<(), Error> {
+///     let req = port.receive()?;
+///     let header = req.header()?;
+///     let message = req.payload()?;
+///     // At this point, you can use the contents of `header`, and the
+///     // bytes in `message`, to construct a response. This will involve
+///     // creating a new header for the response.
+///
+///     // ... do stuff with `message` ...
+///
+/// #   let resp_header = header;
+///     // After calling `reply`, `message` will be inaccessible, so it
+///     // needs to be consumed as far as necessary before this step.
+///     let resp = req.reply(resp_header)?;
+///
+///     let sink = resp.sink()?;
+///     // Now, the computed response can be written to `sink`:
+/// #   let resp_message = [1, 2, 3];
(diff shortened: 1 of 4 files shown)


gkelly:
Hey @hudson-ayers, can you take a look at this too?

src/net.rs
```diff
@@ -0,0 +1,168 @@
...
+    fn receive(&mut self) -> Result<(Address, Header, &mut dyn Read), Error>;
+
+    /// Replies to the request indicated by `addr`.
+    ///
+    /// Calling this function performs sufficient transport-level operations to
+    /// begin a response, before handing off actually composing the payload to
+    /// the caller via a [`Write`].
+    ///
+    /// `addr` must have been previously returned by `recieve()`. Using any
+    /// other value of `addr`, or using it repeatedly, is unspecified behavior.
```
hudson-ayers, line 122: Must the `addr` passed to `reply()` always be the `Address` returned by the most recent call to `receive()`? Or can it be from an older call to `receive()`, so long as that `Address` has not yet been used in a reply? (sorry, I am not familiar with the protocol).

If the former, does it make more sense to leave implementations responsible for storing and reusing the `Address`, rather than leaving it to users of the interface where mistakes are perhaps more likely? For the `InMemPort` example it seems like having reply always use `self.addr` is preferable to requiring the caller to pass the address to `reply()`.

src/net.rs
```diff
@@ -0,0 +1,168 @@
...
+pub trait HostPort {
+    /// Receives an incomming message from a connected host device.
+    ///
+    /// This function will block until a host device indicates that it wishes
+    /// to communicate. At that point, this function will perform enough
+    /// transport-level operations to populate an abstract Manticore header,
+    /// and then hand off parsing to the caller.
+    ///
+    /// Upon success, this function will return:
+    /// - An abstract address, representing the overal request (used with
```
hudson-ayers, line 106:
```suggestion
    /// - An abstract address, representing the overall request (used with
```

src/net.rs
mcy: I actually think you're right. The reason for having `Address` was so that a `HostPort` could have multiple outstanding requests... but in all the likely uses of Manticore, that's probably not going to happen. An eventual `DevicePort` trait, however, will likely need some kind of addressing mechanism.

> The pattern of having the actual payload not be passed to reply() feels a little weird also, but I am guessing there is a reason for letting the user of the interface manually interact with the Write object?

This is to minimize the amount of buffering needed. Rather than requiring the caller to serialize their message somewhere and *then* pipe it into the port, the port is responsible for doing all of that in the most efficient manner possible.

mcy:
I've re-designed the API from the ground up. @hudson-ayers @jrvanwhy could I get your thoughts again?

src/net.rs
```diff
@@ -0,0 +1,331 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Networking interfaces.
+//!
+//! This module provides generic, abstract networking interfaces for working
```
jrvanwhy, line 7: Comment incomplete?

src/net.rs
```diff
@@ -0,0 +1,331 @@
...
+/// A `HostPort` provides all low-level management of incomming and outgoing
+/// connections. An implementation is responsible for packet assembly,
+/// verifying redundancy codes, tracking AEAD tags, and so on, exposing only
+/// an abstract header and a I/O for the message payload itself to Manticore.
+///
+/// Moreover, implementations must be "robust". If a caller, for whatever
+/// reason, drops any of the returned traits on the ground and attempts to
+/// begin a new connection, `HostPort` must also drop any state regarding
+/// that connection on the ground. Because all of the state is tied up with
+/// a mutable reference, implementors can assume that upon calling `recieve()`,
```
jrvanwhy, line 103: s/recieve/receive/g throughout this file.

jrvanwhy:
> I've re-designed the API from the ground up. @hudson-ayers @jrvanwhy could I get your thoughts again?

Overall design seems okay on first pass, but I think it would take me a long time to internalize everything. I feel like there's a better way to do this but I have no idea what it is.

I added a couple of nitpicks.

src/net.rs
```diff
@@ -0,0 +1,331 @@
...
+
+/// The actual guts of an `InMemHost`. This struct is used to implement the two
+/// "connection state" traits used by `HostPort`.
+struct InMemInner<'buf> {
+    rx_header: Option<Header>,
+    rx: &'buf [u8],
+    tx_header: Option<Header>,
+    tx: Cursor<'buf>,
+    finished: bool,
+}
```
hudson-ayers, line 249: I don't understand why this nesting is necessary from the provided examples. The tests compile and pass if `InMemInner` is removed and instead these fields are members of `InMemHost`, and `HostRquest` and `HostReceive` are implemented directly on `InMemHost`.

mcy: I added a comment that explains this indirection.

src/net.rs
```diff
@@ -0,0 +1,331 @@
...
+impl<'buf> InMemHost<'buf> {
+    /// Creates a new `InMemHost`, with the given output buffer for holding
+    /// recieved messages.
+    pub fn new(out: &'buf mut [u8]) -> Self {
+        Self(InMemInner {
+            rx_header: None,
+            rx: &[],
+            tx_header: None,
+            tx: Cursor::new(out),
+            finished: false,
```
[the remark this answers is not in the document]
mcy: No, you're right. I've gone ahead and clarified this.

src/net.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// This function should not be called after calling `reply()`.
+    fn header(&self) -> Result<Header, Error>;
+
+    /// Returns the raw byte stream for the payload of the request.
+    ///
+    /// This function should not be called after calling `reply()`.
+    fn payload(&mut self) -> Result<&mut dyn Read, Error>;
+
+
+    /// Replies to the request indicated by `addr`.
```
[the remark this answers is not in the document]
mcy: Fixed.

mcy:
I've now plugged the new traits into the server framework, and update the tests to match.

mcy:
I imagine that a concrete implementation would look something like this:
```rs
struct MyHost {
  port_state: PortState,
  current_connection: Option<struct {
    header,
    reader_sink,
    reply: Reply,  // Implements HostResponse.
  }>, // Implements HostRequest.
}
```

You do generally have to do the whole nesting dolls business unless you make the struct self-referential. You probably don't want the top leve type to implement HostRequest or HostResponse, just to make it harder for callers to make mistakes there.

gkelly:
Please fix the `receive` spelling before submitting. Thanks Miguel!

PR #37, opened 2020-08-06
[manifest] Add `Containerizer`, a manifest builder.

mcy:
This change adds the necessary functions for actually *making* signed manifests, given a signing key and an in-Rust representation of a manifest we want to sign.

In a followup, I'll add serde support for the FPM, and thread all that together on Oskar's tool.

### DIFF
--- a/src/manifest/container.rs
+++ b/src/manifest/container.rs
@@ -40,9 +40,13 @@
 //! are Manticore-specific.
 
 use crate::crypto::rsa;
+use crate::io;
+use crate::io::cursor::SeekPos;
+use crate::io::Cursor;
 use crate::io::Read as _;
-use crate::manifest::ManifestType;
+use crate::io::Write;
 use crate::manifest::Error;
+use crate::manifest::ManifestType;
 use crate::protocol::wire::WireEnum;
 
 /// Metadata for a [`Container`].
@@ -74,6 +78,12 @@ pub struct Container<'m> {
     metadata: Metadata,
     body: &'m [u8],
 }
+
+/// Offsets for fields within the container header.
+const LEN_OFFSET: usize = 0;
+const TYPE_OFFSET: usize = 2;
+const ID_OFFSET: usize = 4;
+const SIG_LEN_OFFSET: usize = 8;
 
 /// The length of the container header in bytes:
 /// two halves, a word, another half, and two bytes of padding.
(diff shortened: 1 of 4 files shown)


src/manifest/container.rs
```diff
@@ -0,0 +1,490 @@
...
+
+        let mut r = buf; // Use io::Read.
+        let len = r.read_le::<u16>()? as usize;
+        let magic = r.read_le::<u16>()?;
+        let id = r.read_le::<u32>()?;
+        let sig_len = r.read_le::<u16>()? as usize;
+
+        // This length check, combined with the checked arithmetic below,
+        // ensures that none of the slice index operations can panic.
+        if len > buf.len() || len < HEADER_LEN {
```
alphan, line 120: Is the second check necessary? We know that `HEADER_LEN <= buf.len()` from lines 108-110.

src/manifest/container.rs
```diff
@@ -0,0 +1,490 @@
...
+            return Err(Error::OutOfRange);
+        }
+        let rest = &buf[..len][HEADER_LEN..];
```
alphan, line 123: It would be nice to have a comment here that says this takes care of 2 bytes of padding. What do you think?

src/manifest/container.rs
```diff
@@ -0,0 +1,490 @@
...
+
+    #[test]
+    fn parse_manifest() {
+        let keypair =
+            ring::rsa::Keypair::from_pkcs8(testdata::RSA_2048_PRIV_PKCS8)
+                .unwrap();
+        let pub_key = keypair.public();
+        let rsa_builder = ring::rsa::Builder::new();
+        let mut signer = rsa_builder.new_signer(keypair).unwrap();
+        let mut rsa = rsa_builder.new_engine(pub_key).unwrap();
```
alphan, line 363: Can you refactor this part (potentially other lines as well) to a utility function in `mod test` to avoid repetition?

src/manifest/container.rs
```diff
@@ -0,0 +1,490 @@
...
+//! Cerberus manifest containers.
+//!
+//! A "manifest container" is a frame around a Cerberus manifest that provides
+//! certain security properties:
+//! - Containers are signed, so that only the holder of a software signing key
+//!   can create new manifests.
+//! - Containers carry a monotinically increasing "version number", which the
+//!   signing authority should take care to bump for each new manifest.
+//!   Manticore components can refuse to load older manifest versions, as a
+//!   form of playback-attack hardening.
```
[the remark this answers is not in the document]
mcy: Good point!

src/manifest/container.rs
mcy: Good observation.

src/manifest/container.rs
mcy: Done.

src/manifest/container.rs
mcy: Yup!

src/io.rs
```diff
@@ -440,16 +440,101 @@ impl<'a> Cursor<'a> {
...
+    /// cursor.write_bytes(b"barbar")?;
+    ///
+    /// cursor.seek_relative(-6)?;
+    /// cursor.write_bytes(b"foo")?;
+    /// cursor.seek_relative(3)?;
+    /// assert_eq!(cursor.consumed_bytes(), b"foobar");
+    ///
+    /// # Ok::<(), Error>(())
+    /// ```
+    pub fn seek_relative(&mut self, offset: isize) -> Result<(), Error> {
```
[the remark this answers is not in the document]
mcy: That's a good idea. Done.

PR #38, opened 2020-08-07
[io] Explode up `manticore::io` to several files

mcy:
This should make additions to the I/O module easier in the future, such as @osenft's big endian support.


src/io/bit_buf.rs
```diff
@@ -0,0 +1,170 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Provides a way to ergonomically manipilate bytes of queues of bits.
```
alphan, line 5: typo: manipilate

Also, do you mean "bytes as queues of bits" instead of "bytes of queues of bits"?

PR #42, opened 2020-08-24
Add manifest tooling to the cli tool

mcy:
This change also introduces "manifest provenances", which allow us to have operations like `Container::parse()` that produce a distinct type from `Container::parse_and_verify()`.

A signed and JSON FPMs are provided to illustrate what these commands can do.


src/main.rs
```diff
@@ -226,5 +276,99 @@ fn main() {
...
+            let container = match engine {
+                Some(mut engine) => {
+                    match Container::parse_and_verify(&buf, &mut engine) {
+                        Ok(c) => c,
+                        Err(manifest::Error::SignatureFailure) => {
+                            panic!("failed to verify signature")
+                        }
+                        Err(_) => panic!("failed to parse manifest container"),
+                    }
+                    .downgrade()
```
gkelly, line 357: Why does this need to be downgraded?

gkelly (approved):
LGTM. Just one question inline.

src/main.rs
mcy: To make type unification across branches work out. We're going to immediately serialize it so the provenance doesn't matter.

PR #46, opened 2020-10-26
Move `manifest::Container` to a new, buffer-free API

mcy:
This change lays the groundwork for doing "non-buffering" manifest operations, and, eventually, manifest activation.

Manifests are expected to live off-device, and expecting to be able to buffer all of a manifest at any given time is too stringent of a requirement. In a followup, we will expand this support to the FPM.


gkelly (approved):
This took a while to get through, apologies. This looks great, I hadn't caught the signing-the-payload-not-the-signature issue thanks for noticing that.

PR #48, opened 2020-11-12
[flash/mem] Significant changes to how Manticore does memory management

mcy:
This PR is the result of a lot of pain and suffering on my part trying to make the existing Arena and Flash APIs do the right thing for parsing the PFM. I discovered a few things:

- Coupling the lifetimes of allocations in RAM and a Flash object too closely results in some very difficult to resolve lifetime constraints.
- There's no good way for parsing "packed C structs" out of Flash, which is something Cerberus, as a protocol, expects implementors to do.
- The adapters in flash.rs were mostly unhelpful in this regard, since you really want to bind a Flash to a lifetime by using `&'flash Flash`.
- Region/Ptr feel too boilerplatey and hard to use.

The result is this PR, which contains a number of interrelated changes to how Arenas and Flashes interact. The most important once is the elimination of FlashZero, which is replaced with `Flash::direct_read`. This function allows the user to request a zero-copy read, providing their own arena for the allocation if the request can't be fulfilled.

### DIFF
--- a/src/mem/arena.rs
+++ b/src/mem/arena.rs
@@ -8,13 +8,21 @@
 
 use core::cell::Cell;
 use core::marker::PhantomData;
+use core::mem;
 use core::ptr::NonNull;
 use core::slice;
 
 use static_assertions::assert_obj_safe;
 
+use zerocopy::AsBytes;
+use zerocopy::FromBytes;
+use zerocopy::LayoutVerified;
+
+use crate::mem::align_to;
+
 /// An error indicating that an [`Arena`] has run out of allocatable
-/// memory.
+/// memory or that memory that is more aligned than is supported
+/// was requested.
 ///
 /// [`Arena`]: trait.Arena.html
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -29,9 +37,18 @@ pub struct OutOfMemory;
 /// See [`BumpArena`] for an illustration of how the borrow checker is used to
 /// do this safely.
 ///
+/// # Safety
+///
+/// `alloc_aligned()` is required to return memory with certain size and
+/// alignment guarantees. While it itself is a safe function, unsafe code
+/// is permitted to rely on these guarantees.
+///
 /// [`BumpArena`]: struct.BumpArena.html
-pub trait Arena {
-    /// Allocate `len` bytes of memory from this arena.
+pub unsafe trait Arena {
+    /// Allocates `len` bytes of `align`-aligned memory from this arena.
+    ///
+    /// This is a low-level function: prefer instead to use the helpers defined
+    /// in [`ArenaExt`].
     ///
     /// This function may be called multiple times, and the returned slices
     /// will be disjoint.
@@ -41,20 +58,30 @@ pub trait Arena {
     /// As such, "poisoned bits", such as the padding bits of Rust structs,
     /// should never be written to memory returned by this function.
     ///
-    /// Calling `alloc(0)` must never fail.
+    /// Calling `alloc(0, 1)` must never fail. Note that there is no
+    /// requirement that calling `alloc(0, n)` will not return a subslice
+    /// of previously returned memory.
     ///
     /// # Panics
+    ///
+    /// This function will panic if `align` is not a power of two.
     ///
     /// This function is permitted to panic under catastrophic failure
     /// conditions, such as completely running out of program memory.
     /// Implementations must advertize whether they panic.
-    fn alloc(&self, len: usize) -> Result<&mut [u8], OutOfMemory>;
+    ///
+    /// [`ArenaExt`]: trait.ArenaExt.html
+    fn alloc_aligned(
+        &self,
+        len: usize,
+        align: usize,
+    ) -> Result<&mut [u8], OutOfMemory>;
 
     /// Resets this arena, essentially freeing all memory that was given out
     /// and allowing it to be allocated once more.
     ///
     /// Calling this function requires that all slices that were given out by
-    /// `alloc()` are unreachable. Hence, this function must take `self` by
+    /// this arena are unreachable. Hence, this function must take `self` by
     /// unique reference: this ensures that no one else is holding references
     /// to the memory inside.
     ///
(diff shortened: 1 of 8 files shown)


mcy:
@jrvanwhy Would you mind looking over this? Please let me know if something I've done seems odd.

src/mem/arena.rs
```diff
@@ -66,15 +110,80 @@ pub trait Arena {
...
     fn reset(&mut self);
 }
 
 assert_obj_safe!(Arena);
 
+/// Convenience functions for arenas, exposed as a trait.
+///
+/// Note that this trait is implemened for `&impl Arena`, which is the reason
+/// for the slightly odd signature.
+pub trait ArenaExt<'arena> {
```
jrvanwhy, line 126: Instead of using an extension trait, you can put these all into Arena directly as provided methods:

```rust
pub trait Arena {
    fn alloc_aligned(...) -> ...;

    fn alloc<T>(&self) -> ... where ... {
        // implementation here
    }

    etc...
}
```

Then code that currently uses `ArenaExt` would only need to use `Arena`.

src/mem/arena.rs
```diff
@@ -66,15 +110,80 @@ pub trait Arena {
...
+    ) -> Result<&'arena mut T, OutOfMemory> {
+        let bytes =
+            self.alloc_aligned(mem::size_of::<T>(), mem::align_of::<T>())?;
+
+        let lv = LayoutVerified::<_, T>::new(bytes)
+            .expect("alloc_aligned() implemented incorrectly");
+        Ok(lv.into_mut())
+    }
+
+    fn alloc_slice<T: AsBytes + FromBytes + Copy>(
```
jrvanwhy, line 172: It's unclear to me whether `ArenaExt::alloc_slice` will panic, as is required by the trait method documentation (line 154).

src/mem/mod.rs
```diff
@@ -9,3 +9,25 @@ mod arena;
...
+    let (addr, overflow) = addr.overflowing_add(mask);
+    if overflow {
+        return usize::MAX;
+    }
+    addr & !mask
+}
+
+/// Computes the stride of `T`, that is, its size rounded up to its alignment.
+#[inline]
+pub(in crate) fn stride_of<T>() -> usize {
```
jrvanwhy, line 31: `core::mem::size_of::<T>()` is guaranteed to be a multiple of `core::mem::align_of::<T>()` already. See the second sentence of the [size_of docs](https://doc.rust-lang.org/core/mem/fn.size_of.html):

> More specifically, this is the offset in bytes between successive elements in an array with that item type including alignment padding. Thus, for any type `T` and length `n`, `[T; n]` has a size of `n * size_of::<T>()`.

Rust's `size_of` refers to the same concept as Swift calls "stride". If Rust ever has a distinction akin to Swift's "size"/"stride" distinction, then Rust's equivalent to Swift's "size" will need to be called something else. Here's some more background on the topic: https://github.com/rust-lang/rust/issues/33266

src/mem/mod.rs
```diff
@@ -9,3 +9,25 @@ mod arena;
...
+/// `align` must be a power of two; otherwise, the returned value
+/// will be well-defined but unspecified.
+///
+/// This function will always return a value greather than or equal to `addr`.
+#[inline]
+pub(in crate) fn align_to(addr: usize, align: usize) -> usize {
+    let mask = align.saturating_sub(1);
+    let (addr, overflow) = addr.overflowing_add(mask);
+    if overflow {
+        return usize::MAX;
```
jrvanwhy, line 24: You may want to mention that in the case of overflow, this will return an unaligned value. I think returning `usize::MAX` is a bit of a footgun, although my only suggestion is "document it better".

src/mem/mod.rs
```diff
@@ -9,3 +9,25 @@ mod arena;
...
+
+/// Aligns the given address to the alignment for the given type.
+///
+/// `align` must be a power of two; otherwise, the returned value
+/// will be well-defined but unspecified.
+///
+/// This function will always return a value greather than or equal to `addr`.
+#[inline]
+pub(in crate) fn align_to(addr: usize, align: usize) -> usize {
+    let mask = align.saturating_sub(1);
```
jrvanwhy, line 21: I assume this is a saturating subtraction so that if `align` is 0 it won't panic?

It seems to me that using `wrapping_sub` would generate smaller code, and still meet the description "well-defined but unspecified" in the documentation.

src/mem/arena.rs
mcy: Unfortunately, that would make `Arena` not be object-safe, and I'm not excited about using the `where Sized` trick.

jrvanwhy: SGTM

src/mem/arena.rs
```diff
@@ -41,20 +75,30 @@ pub trait Arena {
     /// As such, "poisoned bits", such as the padding bits of Rust structs,
     /// should never be written to memory returned by this function.
     ///
-    /// Calling `alloc(0)` must never fail.
+    /// Calling `alloc(0, 1)` must never fail. Note that there is no
+    /// requirement that calling `alloc(0, n)` will not return a subslice
+    /// of previously returned memory.
```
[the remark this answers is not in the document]
mcy: I fixed it so that it can actually serve really aligned allocations of zero-size.

src/mem/arena.rs
mcy: It can panic due to calls of alloc_aligned.

src/mem/mod.rs
mcy: Wait really? I *swear* that Rust did not promise this. TIL!

mcy:
Jon: I replied to your suggestions. Let me know if you notice any other oddities.

src/mem/arena.rs
jrvanwhy: Sorry, I worded my comment poorly. What is meant was whether it panics must be documented.

src/mem/arena.rs
```diff
@@ -8,18 +8,54 @@
...
 pub struct OutOfMemory;
 
+unsafe impl Arena for OutOfMemory {
+    fn alloc_aligned(
+        &self,
+        len: usize,
+        align: usize,
+    ) -> Result<&mut [u8], OutOfMemory> {
+        assert!(align.is_power_of_two());
+        if len == 0 && align <= mem::align_of::<u128>() {
```
jrvanwhy, line 42: Why limit this to `align <= mem::align_of::<u128>()`? As far as I can tell, the following implementation should be fine:

```rust
assert!(align.is_power_of_two());
if len == 0 {
    Ok(unsafe { &mut *(align as *mut u8) })
} else {
    Err(*self)
}
```

src/mem/arena.rs
```diff
@@ -165,14 +285,63 @@ impl<'arena> Arena for BumpArena<'arena> {
...
+            // to the alignment requirement, if that would land inside the
+            // slice in the first place.
+            let base_addr = self.buf_ptr.as_ptr() as usize;
+            let aligned = align_to(base_addr, align);
+            let misalignment = aligned - base_addr;
+            if misalignment >= self.buf_len {
+                return Err(OutOfMemory);
+            }
+
+            // SAFE: aligned is in-bounds, and zero-length slices have no
```
jrvanwhy, line 333: This can return a slice that is not aligned to `align` if the application's memory is at the top of the available address space (i.e. it extends all the way to `0xFFFFFFFF`).

Suppose that `alloc_aligned` is called with `base_addr == 0xFFFFFFFE`, `buf_len == 2`, `len == 0`, and `align == 2`. Then:

1. We will enter the `if len == 0` branch
1. The `align.is_power_of_two()` check will pass.
1. `align_to` will run its overflow logic and return `usize::MAX`.
1. `misalignment` will be `1`
1. We'll pass the `misalignment >= self.buf_len` check (`1 < 2`)
1. We'll then call `from_raw_parts_mut(0xFFFFFFFF as *mut u8, 0)` and return it to the user.

This doesn't encounter UB inside `alloc_aligned`, but it would return an unaligned slice, which I would expect to cause undefined behavior down the line (where it is presumably casted/transmuted to an `&T` where `T` has alignment `2`).

src/mem/arena.rs
mcy: Excellent observation!

src/mem/arena.rs
mcy: I've changed it so that all zero-sized allocations just call into OutOfMemory.

gkelly (approved):

PR #49, opened 2020-12-03
net: Add a DevicePort implementation

alistair23:
This is a WIP attempt at adding a `DevicePort` implementation to Manticore.

This implementation varies greatly from the current `HostPort` implementation, which is why I'm sending a draft PR to see what others think.

The current `HostPort` implementation seems a little complex and I don't think it will integrate well into libtock-rs and the Tock runtime. For example the `HostPort` `receive`() function is blocking, while libtock-rs is non-blocking.

The `DevicePort` implementation is more geared towards integration with libtock-rs and the Tock ecosystem. I wanted to get feedback on the idea before I spend too much time writing documentation, tests and functionality.

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -17,6 +17,7 @@
 
 #![allow(missing_docs)]
 
+use core::cell::Cell;
 use static_assertions::assert_obj_safe;
 
 use crate::io;
@@ -41,6 +42,8 @@ pub enum Error {
     /// Indicates that some operation was done out of order, such as attempting
     /// to reference part of the request once a reply has begun.
     OutOfOrder,
+    /// The operation timed out.
+    Timeout,
 }
 
 impl From<io::Error> for Error {
@@ -175,6 +178,57 @@ pub trait HostResponse {
     /// Callers should remember to call this function; failing to do so may
     /// result in a response not being sent properly.
     fn finish(&mut self) -> Result<(), Error>;
+}
+
+/// Represents a physical port that can be used to interact with client devices.
+///
+/// This for example is used for a PA RoT to send messages to a AC RoT.
+///
+/// This trait provides a generic mechanism for sending and receiving requests
+/// from devices.
+pub trait DevicePort {
+    /// The send function takes a `header` and `msg` to send to a device.
+    /// The implementation of this function must send the header and msg
+    /// to the device specified by `dest`.
+    ///
+    /// This should block until the operation is complete.
+    ///
+    /// On success it should return nothing.
+    /// On a failure an error is returned.
+    fn send(
+        &mut self,
+        dest: u8,
+        header: Header,
+        msg: &[u8],
+    ) -> Result<(), Error>;
+
+    /// This is a blocking call that should wait for a response or a timeout.
+    ///
+    /// `duration`: Number of milliseconds to wait.
+    ///
+    /// If a response is received in time this should return a success.
+    /// On a time out Error::Timeout is returned.
+    /// Other errors should return a suitable Error as well.
+    fn wait_for_response(&mut self, duration: usize) -> Result<(), Error>;
+
+    /// The `process_response()` function should be called once a reply
+    /// has been received for the device the `send()` operation was sent to.
+    /// `msg` should contain the raw message to be processed.
+    ///
+    /// On success returns the response.
+    fn receive_response(&mut self) -> Result<&mut dyn DeviceResponse, Error>;
+}
+assert_obj_safe!(DevicePort);
+
+/// Provides the "response" half of a transaction with a device.
+///
+/// This for example is used for a PA RoT to send messages to a AC RoT.
+pub trait DeviceResponse {
+    /// Returns the header sent by the device for this response.
+    fn header(&self) -> Result<Header, Error>;
+
+    /// Returns the raw byte stream for the payload of the response.
+    fn payload(&mut self) -> Result<&mut dyn Read, Error>;
 }
 
 /// A simple in-memory [`HostPort`].
@@ -344,5 +398,81 @@ impl HostResponse for InMemInner<'_> {
     fn finish(&mut self) -> Result<(), Error> {
         self.finished = true;
         Ok(())
+    }
+}
+
+/// A simple in-memory [`DevicePort`].
+pub struct InMemDevice<'buf>(InMemDeviceInner<'buf>);
+
+struct InMemDeviceInner<'buf> {
+    rx_header: Cell<Option<Header>>,
+    rx: &'buf [u8],
+    tx_dest: Cell<u8>,
+    tx_header: Cell<Option<Header>>,
+    tx: &'buf mut [u8],
+    finished: Cell<bool>,
+}
+
+impl<'buf> Default for InMemDevice<'buf> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<'buf> InMemDevice<'buf> {
+    /// Creates a new `InMemDeviceInner`, with the given output buffer for holding
+    /// messages to be "transmitted", acting as the final destination for
+    /// replies to this host.
+    pub fn new() -> Self {
+        Self(InMemDeviceInner {
+            rx_header: Cell::new(None),
+            rx: &[],
+            tx_dest: Cell::new(0),
+            tx_header: Cell::new(None),
+            tx: &mut [],
+            finished: Cell::new(false),
+        })
+    }
+
+    pub fn response(&mut self, header: Header, message: &'buf [u8]) {
+        self.0.rx_header.set(Some(header));
+        self.0.rx = message;
+    }
+}
+
+impl<'buf> DevicePort for InMemDevice<'buf> {
+    fn send(
+        &mut self,
+        dest: u8,
+        header: Header,
+        msg: &[u8],
+    ) -> Result<(), Error> {
+        self.0.tx_dest.set(dest);
+        self.0.tx_header.set(Some(header));
+        self.0.tx.copy_from_slice(msg);
+
+        self.0.finished.set(false);
+
+        Ok(())
+    }
+
+    fn wait_for_response(&mut self, _timeout: usize) -> Result<(), Error> {
+        Ok(())
+    }
+
+    fn receive_response(&mut self) -> Result<&mut dyn DeviceResponse, Error> {
+        Ok(&mut self.0)
+    }
+}
+impl DeviceResponse for InMemDeviceInner<'_> {
+    fn header(&self) -> Result<Header, Error> {
+        if !self.finished.get() {
+            return Err(Error::OutOfOrder);
+        }
+        Ok(self.tx_header.get().unwrap())
+    }
+
+    fn payload(&mut self) -> Result<&mut dyn Read, Error> {
+        Ok(&mut self.rx)
     }
 }
--- a/src/net.rs
+++ b/src/net.rs
@@ -451,7 +451,7 @@ impl<'buf> DevicePort for InMemDevice<'buf> {
         self.0.tx_header.set(Some(header));
         self.0.tx.copy_from_slice(msg);
 
-        self.0.finished.set(false);
+        self.0.finished.set(true);
 
         Ok(())
     }
--- a/src/server/pa_rot.rs
+++ b/src/server/pa_rot.rs
@@ -2,7 +2,7 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
-//! A `manticore` "server" for a PA-RoT.
+//! A `manticore` "server" and "client" for a PA-RoT.
 //!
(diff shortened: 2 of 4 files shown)


src/net.rs
```diff
@@ -177,6 +178,44 @@ pub trait HostResponse {
...
+pub trait DevicePort<'a> {
+    /// The send function takes a `header` and `msg` to send to a device.
+    /// The implementation of this function must send the header and msg
+    /// to the device specified by `dest`.
+    ///
+    /// This can be either a blocking of non-blocking call. On success it
+    /// should return the number of bytes sent. If the number of bytes sent
+    /// is 0, but not an error, then the operation is asynchronous. In this
+    /// case there should be an out of band method to indicate the send is
+    /// complete.
```
mcy, line 196: A few thoughts:
- The return value should definitely be an enum, rather than relying on a sentinel.
- What would an "out-of-band" mechanism look like? The problem is that this is probably going to be in the guts of a Manticore state machine, which isn't exactly designed for asynchronicity. How do you envision this trait being used?

src/net.rs
```diff
@@ -177,6 +178,44 @@ pub trait HostResponse {
...
+    /// should be called either after a blocking call completes (`send()`
+    /// returns non zero) or after a non-blocking call completes.
+    fn send_finish(&self) -> Result<(), Error>;
+
+    /// The `process_response()` function should be called once a reply
+    /// has been received for the device the `send()` operation was sent to.
+    /// `msg` should contain the raw message to be processed.
+    ///
+    /// On success returns the header of the response.
+    fn process_response(&self, msg: &[u8]) -> Result<Header, Error>;
```
mcy, line 215: Wait, where does `msg` come from? Shouldn't this trait be responsible for emitting it?

mcy (commented):
Hi Alistair, thanks for taking this on! I've left some comments to try to tease out your design choices for the trait. I've sort of avoided thinking about asynchronicity (mostly because I don't have a good handle on the current state of libtock).

I think the main question to ask is how other parts of Manticore can deal with an asynchronous API correctly. Right now, you're expected to call the request processing function of `PaRot` in a loop, though I was planning to move the loop into the function, since the whole point of HostPort was to make it so `PaRot` could "drive itself".

I'm definitely willing to hear about your overall plans, since that may reveal a better way of doing things asynchronously, though I think we should be clear on what it is *useful* to make asynchronous. For example, under what situations can we have multiple outgoing requests in flight that we can handle in parallel? Recall that we're very tight on memory, so in some cases operations may need to be serialized for this reason (for example, when doing the challenge/response, the certificates are almost certainly going to be too large to have more than one of them in memory at once).

alistair23:
Thanks for the feedback.

I haven't looked too hard into how this would all integrate into Manticore. I wanted to get some early feedback before I spent too much time on this design. I was worried either it has to match `HostPort` or you had already started working on it, and I didn't want to waste too much time on it if that is the case.

Although libtock-rs is async, the yield calls could easily be used to make the calls blocking, so a blocking API will work.

I was picturing a single PA RoT with multiple AC RoTs being able to send multiple requests at once. Let's say that some AC RoTs are very slow it seems a shame to stall the entire process waiting on them. Like you mention though, the PA RoT might be too limited to be able to handle this situation anyway, so it might not be beneficial though.

As for the loop, we could make a blocking call maybe allowing both? I'll keep digging and see what seems to work.

src/net.rs
alistair23: Ah, I was going for a callback style here, but I don't think that will work. I'll fix this

src/net.rs
alistair23: I have improved this, I'm going to mark this as resolved as it has been removed.

alistair23:
I pushed an update that more closely matches the current `HostPort` but should also be a little more libtock-rs friendly.

alistair23:
I have marked this as ready to review. I have an integration with libtock-rs that is able to query and decode AC firmware version via MCTP.

mcy:
Apparently I forgot to look at this, sorry! I'll TAL after the holidays. In the future, if I don't reply, please make sure to hit the little "recycle" button in the reviewers list, so that I don't miss things. (Or, ping me on Slack et. al.)

alistair23:
Ping! I don't see a recycle button to click.

src/net.rs
```diff
@@ -177,6 +180,57 @@ pub trait HostResponse {
...
+    /// The `process_response()` function should be called once a reply
+    /// has been received for the device the `send()` operation was sent to.
+    /// `msg` should contain the raw message to be processed.
+    ///
+    /// On success returns the header of the response.
+    fn receive_response(&mut self) -> Result<&mut dyn DeviceResponse, Error>;
+}
+assert_obj_safe!(DevicePort);
+
+pub trait DeviceResponse {
```
mcy, line 222: Missing a comment here.

src/net.rs
```diff
@@ -177,6 +180,57 @@ pub trait HostResponse {
...
+    fn send(
+        &mut self,
+        dest: u8,
+        header: Header,
+        msg: &'a [u8],
+    ) -> Result<usize, Error>;
+
+    /// This is a blocking call that should wait for a response or a timeout.
+    ///
+    /// `timeout`: Number of milliseconds to wait.
```
mcy, line 207: Can this be a `Duration` instead?

Also, hmm, I wonder if the contract should specify that the timeout is best effort... it might be the case that the underlying hardware just doesn't support a timeout. Thoughts?

src/net.rs
```diff
@@ -177,6 +180,57 @@ pub trait HostResponse {
...
+    ///
+    /// If a response is received in time this should return a success.
+    /// On a time out or error a error should be returned.
```
mcy, line 210: Nit: specify that `Error::Timeout` is required for a timeout? It's obvious but worth clarifying.

src/net.rs
```diff
@@ -177,6 +180,57 @@ pub trait HostResponse {
...
+    fn wait_for_response(&mut self, timeout: usize) -> Result<(), Error>;
+
+    /// The `process_response()` function should be called once a reply
+    /// has been received for the device the `send()` operation was sent to.
+    /// `msg` should contain the raw message to be processed.
+    ///
+    /// On success returns the header of the response.
```
mcy, line 217: This comment looks stale?

src/net.rs
```diff
@@ -177,6 +180,57 @@ pub trait HostResponse {
...
+    fn receive_response(&mut self) -> Result<&mut dyn DeviceResponse, Error>;
+}
+assert_obj_safe!(DevicePort);
+
+pub trait DeviceResponse {
+    /// Returns the header sent by the host for this request.
+    ///
+    /// This function should not be called after calling `reply()`.
```
mcy, line 225: I think you might have missed updating the comments in this trait.

src/net.rs
```diff
@@ -177,6 +180,57 @@ pub trait HostResponse {
...
+/// This trait provides a generic mechanism for sending and receiving requests
+/// from devices.
+pub trait DevicePort<'a> {
+    /// The send function takes a `header` and `msg` to send to a device.
+    /// The implementation of this function must send the header and msg
+    /// to the device specified by `dest`.
+    ///
+    /// This should block until the operation is complete.
+    ///
+    /// On success it should return the number of bytes sent.
```
mcy, line 196: What's a caller expected to do with this information? Seems like it might make more sense to return a `&mut dyn io::Write` instead of accepting a slice (this means, conveniently, that you don't need to buffer the message you're planning to send).

src/net.rs
```diff
@@ -177,6 +180,57 @@ pub trait HostResponse {
...
+    /// On a failure an error is returned.
+    fn send(
+        &mut self,
+        dest: u8,
```
mcy, line 200: I wonder if this should be a trait parameter... let's leave it for now.

src/net.rs
```diff
@@ -177,6 +180,57 @@ pub trait HostResponse {
     fn finish(&mut self) -> Result<(), Error>;
 }
 
+/// Represents a physical port that can be used to interact with client devices.
+///
+/// This for example is used for a PA RoT to send messages to a AC RoT.
+///
+/// This trait provides a generic mechanism for sending and receiving requests
+/// from devices.
+pub trait DevicePort<'a> {
```
mcy, line 189: What does this lifetime indicate? I suspect you don't *actually* want it.

mcy (commented):
Hi Alistair! Sorry I took so long to review this, I'll try to pay more attention next time.

I've left two kinds of comments:
- Thoughts on details of the design. Overall, this is a good direction, but I have a few places where I feel uncertain.
- Nits on documentation. You'll find I'm a bit of a stickler about having complete docs. =)

Re the "recycle" button, you didn't see it because I pressed it. Now that I've done a review, it should show up next to my names in the "reviewers" pane. Of course, if I'm taking to long anyways, please reach out if I'm blocking you.

src/net.rs
alistair23: I looked at returning an `io::Write` and using that, but it uses a large amount of stack space that I was trying to avoid. It also seems overally complex. I would prefer to just send the data from this function.

src/net.rs
alistair23: It's required for borrowing the buffer.

alistair23:
I have addressed some comments and replied to others.

src/net.rs
mcy: Hmm, I don't follow why it uses "a lot" of stack space? Can you quantify?

While it is simpler to just send the data this way, it requires the sender to buffer up the encoded message, which may require using a lot of memory to do so. Avoiding this is one of Manticore's design goals (which has required significant complexity elsewhere). If you'd like we can merge this as-is, but I will probably wind up having to refactor it later.

src/net.rs
mcy: Why can't you accept `&[u8]` rather than giving it a fixed lifetime?

src/server/pa_rot.rs
```diff
@@ -245,6 +259,55 @@ mod test {
...
     }
 
+    fn simulate_response<'a, C: protocol::Command<'a>, A: Arena>(
+        scratch_space: &'a mut [u8],
+        arena: &'a mut A,
+        _server: &mut PaRot<fake::Identity, fake::Reset, ring::rsa::Builder>,
+        request: C::Req,
+    ) -> Result<C::Resp, Error> {
+        let len = scratch_space.len();
+        let (req_scratch, _port_scratch) = scratch_space.split_at_mut(len / 2);
```
mcy, line 269: Nit: `_port_scratch` => `_`?

src/server/pa_rot.rs
```diff
@@ -245,6 +259,55 @@ mod test {
         Ok(resp_val)
     }
 
+    fn simulate_response<'a, C: protocol::Command<'a>, A: Arena>(
+        scratch_space: &'a mut [u8],
+        arena: &'a mut A,
+        _server: &mut PaRot<fake::Identity, fake::Reset, ring::rsa::Builder>,
```
mcy, line 265: Is there a particular reason you're taking this as an argument?

src/net.rs
alistair23: Yeah, so the current libtock-rs implementation uses 0x800 bytes of stack. When I changed this to returning an `io::Write` the stack was exceeded (and generated a fault), if I bump it up to 0x1000 it works again.

I can get more concrete numbers if you would like.

I'm not sure I understand the advantage of using `io::Write`. It adds a lot of complexity and is there a real case where you can get away without buffering the data? What hardware interface are you imaging?

src/net.rs
alistair23: Rust complains that the lifetimes don't match then for the implementation of `InMemDevice`

This is similar to the existing `request()` function: https://github.com/lowRISC/manticore/blob/master/src/net.rs#L285

src/server/pa_rot.rs
alistair23: I prefer keeping the name here so it's more obvious what it is for future use.

src/net.rs
mcy: Let's just leave it like this for now. I can't promise I'll keep it like this indefinitely, though I still think that returning a `usize` is wrong and it should just block.

src/net.rs
mcy: This is only necessary if you're going to take ownership of the buffer and save it, which doesn't seem to make a lot of sense for what is morally a blocking `write()` syscall? I would have expected the in-memory version to perform a copy (which it will have to do anyway if I find that this needs to take `Write` instead of a buffer.)

mcy (commented):
Allistair: replied to a couple of comments. I think we're converging on something good here, and I'd rather unblock you than keep bikeshedding whether we should use a Write or not (though I can't promise I won't need to change this in the future).

I think my remaining comments should be easy to resolve at this point, and then I'll go ahead and merge.

alistair23:
Ok, should be good to go now.

mcy (approved):

PR #50, opened 2020-12-03
Run cargo clippy as part of the CI

alistair23:


src/io/bit_buf.rs
```diff
@@ -119,6 +124,12 @@ impl BitBuf {
     }
 }
 
+impl Default for io::bit_buf::BitBuf {
```
mcy, line 127: Shouldn't this be just `BitBuf`?

mcy (approved):
Was meaning to get to this for a while. Thanks Alistair!

src/io/bit_buf.rs
alistair23: Fixed.

alistair23:
Ping!

mcy (approved):
LG. =)

PR #51, opened 2021-01-06
[manifest] Add PFM support (for real this time!)

mcy:
This is the first change that adds basic PFM support: parsing a PFM's TOC, and extracting the simples of the PFM elements (the Platform ID). It also includes a simple test that makes sure the simplest case works.

### DIFF
--- a/src/manifest/pfm.rs
+++ b/src/manifest/pfm.rs
@@ -1,0 +1,497 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! The Platform Firmware Manifest (PFM)
+//!
+//! A PFM is a computer-readable policy document describing firmware which is
+//! allowed to run on a device. Practically, this means that a PFM describes
+//! an allowed configuration of an external storage device, such as a SPI
+//! flash device. It also carries some additional metadata about the platform
+//! itself.
+//!
+//! The PFM is designed to be readable by "random access". It is encoded as a
+//! ["table of contents"](struct.Toc.html) that describes various blobs that
+//! follow it; this table of contents makes it possible to pull different parts
+//! of the PFM out in arbitrary order. The PFM also contains hashes for
+//! ensuring that all data read from flash is protected by the original
+//! signature check.
+//!
+//! The [`Pfm`] type is the entry-point for this module.
+//!
+//! [`Pfm`]: struct.Pfm.html
+
+use core::mem;
+
+use zerocopy::AsBytes;
+use zerocopy::FromBytes;
+
+use crate::crypto::sha256;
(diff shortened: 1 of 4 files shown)


src/manifest/pfm.rs
```diff
@@ -0,0 +1,497 @@
...
+    ///
+    /// This value is measured from the start of the PFM's manifest frame,
+    /// rather than from the start of the PFM's table of contents.
+    pub fn offset(&self) -> u16 {
+        self.offset as _
+    }
+
+    /// Returns the length, in bytes, of this entry's element.
+    pub fn len(&self) -> u16 {
+        self.len as _
```
gkelly, line 120: These are both casts from u16 to u16? Did you want the return type to be usize for iteration's sake?

src/manifest/pfm.rs
```diff
@@ -0,0 +1,497 @@
...
+    hashes: &'toc [sha256::Digest],
+}
+
+impl<'toc> Toc<'toc> {
+    /// Checks that all invariants of this `Toc` type hold:
+    /// - Every pointer to a hash is in-bounds.
+    /// - Every pointer to a parent is in-bounds.
+    ///
+    /// Returns true if the invariants are upheld.
+    fn verify_invariants(&self) -> bool {
```
gkelly, line 186: Is there a requirement that the PFM be acyclic?

src/manifest/pfm.rs
```diff
@@ -0,0 +1,497 @@
...
+    pub fn entries_of_type(
+        &self,
+        ty: ElementType,
+    ) -> impl Iterator<
+        Item = (usize, &'toc TocEntry, Option<&'toc sha256::Digest>),
+    > + '_ {
+        self.entries()
+            .enumerate()
+            .map(|(i, (e, h))| (i, e, h))
+            .filter(move |(_, e, _)| e.element_type() == Some(ty))
```
gkelly, line 253: I was going to suggest filtering before mapping, but given it's a type shuffle I don't think it matters, right?

src/manifest/pfm.rs
```diff
@@ -0,0 +1,497 @@
...
+    }
+
+    /// Returns the first entry in this `Toc` of a particular type, if any.
+    pub fn entry_of_type(
```
gkelly, line 257: `first_entry_of_type`?

gkelly (approved):

src/manifest/pfm.rs
mcy: One would hope that that map operation is a no-op, given that all three elements have the same size and alignment.

src/manifest/pfm.rs
mcy: Not currently! Probably something to bring to Microsoft later on.

src/manifest/pfm.rs
mcy: Yeah I think that was supposed to be usize.

PR #52, opened 2021-01-11
[crypto] Make crypto traits return a "known error type"

mcy:
The main benefit of this change is removing the inability to implement `From<crypto::Error>` for e.g. `manifest::Error`. This will allow us to reduce the boilerplate around using crypto functions.

Additionally, this gives us the ability to make Manticore understand some crypto errors in the future, such as a "unsupported key" error in RSA et. al.


src/crypto/rsa.rs
```diff
@@ -84,16 +84,41 @@ pub trait Keypair {
...
+/// This type serves as a combination of built-in error types known to
+/// Manticore, plus a "custom error" component for surfacing
+/// implementation-specific errors that Manticore can treat as a black box.
+///
+/// This type has the benefit that, unlike a pure associated type, `From`
+/// implementations for error-handling can be implemented on it.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum Error<E = ()> {
+    /// The "custom" error type, which is treated by Manticore as a black box.
+    Custom(E),
```
gkelly, line 98: I wonder if this enum ever gets any other elements. I'm fine to leave it like this, but part of me thinks that it won't ever be extended.

gkelly (approved):

src/crypto/rsa.rs
mcy: I'm thinking it *might*. I can imagine we might decide to add a "retry" variant or a "key size not supported" variant. I mostly chose an enum for flexibility. Worst case we can turn this into a tuple struct and not think about it too hard.

PR #56, opened 2021-01-29
Rewrite the Manifest parsing framework

mcy:
This change completely refactors how manifests are parsed at the TOC level.

This change includes:
- Integrated TOC support in `Container`.
- New `manticore::manifest::owned` module for heap-allocated manifest trees, for use by tooling.
- PFM support in the manticore tool.
- FPM removal.

This change does not make the parser completely compliant with Cerberus; there's a few holes here and there that I intend to plug in a series of smaller followups. These include:
- Questionable handling of format versions.
- Correctly ignoring element types we don't understand.
- The byte layout of the TOC entries.
- Correct handling of various element flags.
- Support for non-SHA-256 algorithms.
- Support for ECC.


src/manifest/container.rs
```diff
@@ -75,6 +79,238 @@ pub struct Metadata {
...
+///
+/// A TOC entry is encoded exactly the same way as this struct is laid
+/// out, byte-for-byte.
+///
+/// See [`Toc`](struct.Toc.html).
+pub struct TocEntry<'entry, 'toc, M> {
+    toc: &'entry Toc<'toc, M>,
+    // Invariant: index is always a valid index into `toc.entries`.
+    index: usize,
+    _ph: PhantomData<fn() -> M>,
```
gkelly, line 92: With your `toc` taking type parameter `M` is this needed?

src/manifest/container.rs
```diff
@@ -75,6 +79,238 @@ pub struct Metadata {
...
+    }
+
+    /// Returns this entry's parent, if it has one.
+    pub fn parent(self) -> Option<Self> {
+        match self.raw().parent_idx {
+            0xff => None,
+            x => Some(
+                self.toc
+                    .entry(x as usize)
+                    .expect("previously verified by check_invariants()"),
```
gkelly, line 133: I feel like this is an `unreachable!()`, and the string is just bloating the binary. I don't think I care for now, though.

src/manifest/container.rs
```diff
@@ -75,6 +79,238 @@ pub struct Metadata {
...
+}
+
+impl<'toc, M: Manifest> Toc<'toc, M> {
+    /// Checks that all invariants of this `Toc` type hold:
+    /// - Every pointer to a hash is in-bounds.
+    /// - Every pointer to a parent is in-bounds.
+    /// - Every type/version pair is well-known.
+    ///
+    /// Returns true if the invariants are upheld.
+    #[allow(unused)]
```
gkelly, line 216: Why do you need to `#[allow(unused)]` here, shouldn't this be being used to check the invariants?

src/hardware/flash.rs
```diff
@@ -519,7 +519,23 @@ impl<F: Flash> io::Write for FlashIo<F> {
...
+    #[cfg_attr(
+        feature = "serde",
+        serde(deserialize_with = "crate::serde::de_radix")
+    )]
+    #[cfg_attr(
+        all(feature = "serde", feature = "std"),
+        serde(serialize_with = "crate::serde::se_hex")
+    )]
     pub offset: u32,
+    #[cfg_attr(
```
gkelly, line 531: nit: Maybe a newline above here just to help readability on these blocks.

src/manifest/owned/mod.rs
```diff
@@ -0,0 +1,412 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Owned manifest containers.
+//!
+//! TODO(mcyoung): Document this module. This should be done in a followup
```
gkelly: `TODO(issue#)`.

src/manifest/owned/mod.rs
```diff
@@ -0,0 +1,412 @@
...
+
+    /// Returns the element type of a specific element.
+    fn element_type(&self) -> Self::ElementType;
+
+    /// Attempts to construct an `Element` of the given type from encoded
+    /// `bytes`.
+    fn from_bytes(ty: Self::ElementType, bytes: &[u8]) -> Result<Self, Error>;
+
+    /// Attempts to encode this `Element` into bytes, using the given
+    /// padding byte as "filler".
```
gkelly, line 65: My eyebrows are raised at this filler byte, I continue reading with great interest.

src/manifest/owned/mod.rs
```diff
@@ -0,0 +1,412 @@
...
+/// [`manticore::manifest::pfm`]: ../pfm/index.html
+pub type Pfm = Container<self::pfm::Element>;
+
+/// A heap-allocated Cerberus manifest, represented as a tree structure.
+///
+/// Prefer to access this type through one of the provided type aliases,
+/// instead:
+/// - Platform Firmware Manifest: [`Pfm`](type.Pfm.html)
+/// - Component Firmware Manifest: NYI
+/// - Platform Configuration Descriptor: NTI
```
gkelly, line 82: NYI

src/manifest/owned/mod.rs
```diff
@@ -0,0 +1,412 @@
...
+    pub bad_hashes: Vec<usize>,
+}
+
+impl<E: Element> Container<E> {
+    /// Parses a `Container` out of `bytes`.
+    ///
+    /// Note that if a cryptographic operation fails, `Ok` will still be returned;
+    /// the resulting value will indicate which operations failed in addition to
+    /// containing the parsed container.
+    pub fn parse(
```
gkelly, line 147: How come you can't reuse the parsing code? Having N>1 parsers can lead to a lot of issues, many of which have security implications.

src/manifest/owned/pfm.rs
```diff
@@ -0,0 +1,586 @@
...
+    fn element_type(&self) -> ElementType {
+        match self {
+            Self::FlashDevice { .. } => ElementType::FlashDevice,
+            Self::AllowableFw { .. } => ElementType::AllowableFw,
+            Self::FwVersion { .. } => ElementType::FwVersion,
+            Self::PlatformId { .. } => ElementType::PlatformId,
+        }
+    }
+
+    fn from_bytes(ty: ElementType, mut bytes: &[u8]) -> Result<Self, Error> {
```
gkelly, line 146: Same question re parsers.

src/serde.rs
```diff
@@ -0,0 +1,132 @@
...
+                let int = if s.starts_with("0b") || s.starts_with("0B") {
+                    $ty::from_str_radix(&s[2..], 0b10)
+                }
+                else if s.starts_with("0o") || s.starts_with("0O") {
+                    $ty::from_str_radix(&s[2..], 0o10)
+                }
+                else if s.starts_with("0x") || s.starts_with("0X") {
+                    $ty::from_str_radix(&s[2..], 0x10)
+                } else {
+                    $ty::from_str_radix(s, 10)
```
gkelly, line 73: 2, 8, 16 would be significantly more readable.

src/manifest/container.rs
```diff
@@ -371,69 +364,125 @@ impl<F: Flash> Container<F, provenance::Signed> {
...
+        // Container header.
         let len = r.read_le::<u16>()? as usize;
         let magic = r.read_le::<u16>()?;
         let id = r.read_le::<u32>()?;
         let sig_len = r.read_le::<u16>()? as usize;
-
-        // This length check, combined with the checked arithmetic below,
-        // ensures that none of the slice index operations can panic.
-        if len > flash_len {
+        // FIXME: Manticore currently ignores this value.
```
gkelly, line 394: `TODO(issue#)`

src/manifest/container.rs
mcy: Nope.

src/manifest/container.rs
mcy: I think the right thing to do is, if we find that such strings bloat the binary, add some kind of `unwrap!` macro where we can disable the error messages with a feature flag. Let's do that once we start running into size problems.

src/manifest/container.rs
mcy: Nope.

src/serde.rs
mcy: Too clever huh?

src/manifest/owned/mod.rs
mcy: This was actually done in a followup commit. =)

src/manifest/owned/mod.rs
mcy: This is currently a stopgap, due to the way the commits are layered. The idea is actually to use the "unowned" Container/PFM parser to walk an encoded manifest and dump it into the heap, but that requires re-doing the API for Container/PFM that would have made this PR even bigger than it has to be[1]. Look out for a PR that will eliminate this parser.

src/manifest/owned/pfm.rs
mcy: (See above.)

src/manifest/owned/mod.rs
mcy: Right now this exists to allow padding with `0xff` or `0x00`. Conversations with Chris indicate that this is incorrect and `0x00` is the only correct value, so this parameter will disappear in a followup.

src/manifest/owned/mod.rs
gkelly: Got it, thanks!

gkelly (approved):
LGTM!

PR #62, opened 2021-02-02
Unify the two copies of the Manifest/PFM parser into one

mcy:
This is a followup for #56

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -187,13 +187,31 @@ pub trait Manifest {
 ///
 /// These types are only really intended to be used as type parameters.
 pub mod provenance {
+    /// A provenance. This trait can be used to write code that is generic
+    /// on provenances but which might choose to skip certain operations for
+    /// the `Adhoc` provenenace.
+    pub trait Provenance {
+        /// Whether this provenance represents an authenticated source,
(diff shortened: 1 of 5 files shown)


src/manifest/mod.rs
```diff
@@ -187,13 +187,30 @@ pub trait Manifest {
 ///
 /// These types are only really intended to be used as type parameters.
 pub mod provenance {
+    /// A provenance. This trait can be used to write code that is generic
+    /// on provenances but which might choose to skip certain operations for
+    /// the `Adhoc` provenenace.
+    pub trait Provenance {
+        /// Whether this provenance is "trusted", i.e., whether it represents
+        /// a continuous chain of trust up to a signature.
+        const TRUSTED: bool;
```
gkelly, line 196: I'm not 100% on board with `TRUSTED` as the designator here. I _trust_ that an adhoc `Provenance` comes from nowhere. Possible followup but I'm OK with this for now.

gkelly (approved):
`cargo fmt` and you're set.

src/manifest/mod.rs
mcy: Changed to `AUTHENTICATED` per our discussion.

PR #63, opened 2021-02-04
[manifest] Improve the quality of errors emitted by the `manifest` module.

mcy:
Unfortunately, this does not include `thiserror` support, because that crate requires `std`.

### DIFF
--- a/src/manifest/pfm.rs
+++ b/src/manifest/pfm.rs
@@ -113,9 +113,6 @@ where
             Some(x) => x,
             None => return Ok(None),
         };
-        if entry.region().len < 4 {
-            return Err(Error::OutOfRange);
-        }
 
         let data =
             self.container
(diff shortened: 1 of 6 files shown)


mcy:
@cfrantz I feel like this is a good first PR for you to review. TAL when you get a chance, and feel free to reach out w/ questions.

src/manifest/pfm.rs
```diff
@@ -412,88 +448,120 @@ where
...
         // size, which have already been explicitly checked above.
         let rw_regions = LayoutVerified::<_, [RwRegion]>::new_slice(rw_bytes)
             .unwrap()
             .into_slice();
+
         for rw in rw_regions {
-            if rw.start_addr > rw.end_addr {
-                return Err(Error::OutOfRange);
-            }
+            Region::from_start_and_limit(rw.start_addr, rw.end_addr).ok_or(
```
cfrantz, line 502: Creates and discards a new Region just for the error checking properties of `from_start_and_limit`?

More clear as `let _ = Region::from_start_and_limit(...).ok_or(...)?;`

(Not sure on what is idiomatic Rust in this case).

cfrantz (approved):

src/manifest/pfm.rs
mcy: Yup, that makes sense. I don't think the fact that it was called just for error-checking is obvious.

PR #64, opened 2021-02-04
Improve correctness of manifest parsing/encoding

mcy:
This change brings us to almost perfect parity with Microsoft's implementation; the commit series culminates in a change that adds a test against a PFM generated using Microsoft's own test tooling.

However, we still differ in how we verify the signature. Resolving this issue is beyond the scope of this change.


src/manifest/container.rs
```diff
@@ -144,9 +148,20 @@ impl<'entry, 'toc, M: Manifest> TocEntry<'entry, 'toc, M> {
 
     /// Returns an iterator over all of this entry's children.
     pub fn children(self) -> impl Iterator<Item = TocEntry<'entry, 'toc, M>> {
-        self.toc
-            .entries()
-            .filter(move |e| self.index == e.raw().parent_type as usize)
+        let mut index = self.index() + 1;
+        std::iter::from_fn(move || loop {
```
gkelly, line 151: `#![no_std]`

src/manifest/container.rs
```diff
@@ -298,6 +286,18 @@ impl<'toc, M: Manifest> Toc<'toc, M> {
             index: i,
         })
     }
+
+    /// Returns the first element of the given type in this `Toc`.
+    ///
+    /// This function should be used to discover "singleton" entries, entries
+    /// such that only the very first apparence thereof in a toc is used, with
```
gkelly, line 293: `apparence` -> `appearance`

src/manifest/pfm.rs
```diff
@@ -889,4 +890,65 @@ mod test {
...
+
+    #[test]
+    fn baked_pfm1() {
+        let sha = ring::sha256::Builder::new();
+        let (mut rsa, mut signer) = test_rsa();
+        let mut arena = vec![0; 1024];
+        let arena = BumpArena::new(&mut arena);
+
+        let bytes = Ram(testdata::PFM_RSA1);
+        // NOTE: Manticore and MS Cerberus currently have inconsistent signing
```
gkelly, line 902: Remove "MS" here.

gkelly (approved):

src/crypto/testdata/rsa_2048_private_key.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN RSA PRIVATE KEY-----
```
cfrantz, line 1: Are there safety checks around submitting private keys into the repo?

Based on the path, this key is for testing, but in the past, keys have been exposed through error.

src/manifest/pfm.rs
```diff
@@ -897,4 +890,65 @@ mod test {
...
+        assert_eq!(
+            fw.version(),
+            (Region::new(0x12345, 7), b"Testing".as_ref())
+        );
+        assert_eq!(fw.rw_count(), 1);
+        assert_eq!(fw.image_count(), 1);
+
+        let rws = fw.rw_regions().collect::<Vec<_>>();
+        assert_eq!(rws.len(), 1);
+        assert_eq!(rws[0].region(), Region::new(0x200_0000, 0x200_0000 - 1));
```
cfrantz, line 945: Is the second param here a length or something more like a limit (here and on L951).

The current code for Region seems to imply this is a length.

src/manifest/pfm.rs
```diff
@@ -897,4 +890,65 @@ mod test {
...
+        // schemes, so we disable signature verification (for now).
+        /*let container = Container::parse_and_verify(
+            &bytes,
+            &sha,
+            &mut rsa,
+            &arena,
+            &OutOfMemory,
+        )
+        .unwrap();*/
+        let container = Container::parse(&bytes, &arena).unwrap();
```
cfrantz, line 912: A question about idiomatic rust: According to [the docs](https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#tests-and-), tests are permitted to return `Result<()>` so you can used `?` in place of `unwrap()`.

Do we allow that in our codebase?

src/crypto/testdata/rsa_2048_private_key.pem
mcy: We do not. =/

If you have presubmit suggestions I'm happy to hear them. Can you file an issue?

src/manifest/pfm.rs
mcy: Can you check that this is still an issue?

mcy: That's right; previously, the PFM was being parsed incorrectly and the PR I merged the other day fixed this (due to confusion between lengths/limits). I added the -1 to make it pass, but now I can remove it.

src/manifest/pfm.rs
mcy: So, I believe that's a modern style; I've never done that but that's because I've been writing this language for a while, so I missed when this was added.

I think this is a good idea. Do you think we should file an issue to update the tests? I think it's a good task to leave to someone less-familiar with the codebase so they can get a feel for what all the names of everything is.

src/crypto/testdata/rsa_2048_private_key.pem
cfrantz: https://github.com/lowRISC/opentitan/issues/5843

src/manifest/pfm.rs
cfrantz: SGTM.  Do we also have rust styleguide which would inform about such things?

https://github.com/lowRISC/opentitan/issues/5844

src/manifest/pfm.rs
```diff
@@ -967,4 +966,65 @@ mod test {
...
     }
+
+    #[test]
+    fn baked_pfm1() {
+        let sha = ring::sha256::Builder::new();
+        let (mut rsa, mut signer) = test_rsa();
+        let mut arena = vec![0; 1024];
+        let arena = BumpArena::new(&mut arena);
+
+        let bytes = Ram(testdata::PFM_RSA1);
```
cfrantz, line 977: Since this is a generated binary resource, maybe some comments which describe what one should expect?

Doesn't necessarily have to be here - perhaps in testdata/mod.rs next to the declaration of PFM_RSA1.

cfrantz (approved):

PR #65, opened 2021-04-06
[manifest] Add an intermediate type to represent all PFMs as a whole

mcy:
This change has two benefits:
- It removes proliferation of repeated type parameters all over the place,
  since we can now write `Pfm` instead of `ParsedPfm<'all, 'kinds, Of,
  Stuff>`.
- It makes it possible to have code generic over all Pfms that can
  specialize a ParsedPfm for different Flash types and whatnot with
  minimal threading of type parameters around.

This will be particularly useful when we add the manifest manager, which needs to be able to deal with manifests of fixed type but different underlying storage.

### DIFF
--- a/src/manifest/owned/pfm.rs
+++ b/src/manifest/owned/pfm.rs
@@ -259,14 +259,19 @@ impl owned::Element for Element {
 }
 
 impl<'f, F: 'f + Flash> owned::FromUnowned<'f, F> for Element {
-    type Unowned = manifest::pfm::Pfm<'f, F, provenance::Adhoc>;
+    type Manifest = manifest::pfm::Pfm;
 
     fn from_container(
-        container: manifest::Container<'f, Self::Unowned, F, provenance::Adhoc>,
+        container: manifest::Container<
+            'f,
+            Self::Manifest,
+            F,
+            provenance::Adhoc,
+        >,
     ) -> Result<Vec<owned::Node<Self>>, Error> {
         let mut arena = vec![0; 2048];
         let mut arena = BumpArena::new(&mut arena);
-        let pfm = manifest::pfm::Pfm::new(container);
+        let pfm = manifest::pfm::ParsedPfm::new(container);
         let sha = RingSha::new();
         let mut nodes = Vec::new();
 
--- a/src/manifest/pfm.rs
+++ b/src/manifest/pfm.rs
@@ -17,9 +17,9 @@
 //! ensuring that all data read from flash is protected by the original
 //! signature check.
 //!
-//! The [`Pfm`] type is the entry-point for this module.
+//! The [`ParsedPfm`] type is the entry-point for this module.
 //!
-//! [`Pfm`]: struct.Pfm.html
+//! [`ParsedPfm`]: struct.ParsedPfm.html
 
 use core::mem;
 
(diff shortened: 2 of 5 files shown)


src/manifest/owned/pfm.rs
```diff
@@ -259,14 +259,19 @@ impl owned::Element for Element {
...
+    type Manifest = manifest::pfm::Pfm;
 
     fn from_container(
-        container: manifest::Container<'f, Self::Unowned, F, provenance::Adhoc>,
+        container: manifest::Container<
+            'f,
+            Self::Manifest,
+            F,
+            provenance::Adhoc,
+        >,
```
jrvanwhy, line 270: This isn't an actionable comment, but I don't like what `rustfmt` did here.

jrvanwhy (approved):

src/manifest/pfm.rs
```diff
@@ -801,7 +856,7 @@ mod test {
         let (mut rsa, mut signer) = test_rsa();
 
         #[rustfmt::skip]
-        let pfm: owned::Pfm = from_str(r#"{
+        let pfm: owned::ParsedPfm = from_str(r#"{
```
jrvanwhy, line 859: Oops, I hit approve before I remembered that this doesn't build. Was this supposed to be `crate::manifest::pfm::ParsedPfm`?

jrvanwhy (changes requested):

src/manifest/owned/pfm.rs
mcy: Same. I actually tried to throw a #[rustfmt::skip]` on the `foo: Stuff<...>` binding and it was ignored. I can clean this up and rustfmt::skip the whole function decl if you'd prefer.

src/manifest/pfm.rs
mcy: Embarassing!

No idea how that slipped through.

src/manifest/owned/pfm.rs
jrvanwhy: No, let's leave it as is rather than `rustfmt::skip`-ing a larger chunk of code.

jrvanwhy (approved):

src/manifest/pfm.rs
```diff
@@ -82,14 +91,60 @@ impl<F, P> Manifest for Pfm<'_, F, P> {
...
+
+    fn copy_to<F2: Flash>(
+        manifest: &Self::Parsed,
+        dest: &mut F2,
+    ) -> Result<(), Error> {
+        let src = manifest.container.flash();
+        let len = src.size()? as usize;
+        let mut bytes_left = len;
+
+        let mut buf = [0; 32];
```
cfrantz, line 111: nit: `[0u8; 32]` to be more obvious this is a byte buffer.

nit2: 32 bytes is pretty small for flash - generally a "page" in flash is 256 bytes (I see your flash trait specifically says that `program` is expected to optimize the actual writing to flash.  Given the differences between sectors (typ. 4K) and pages (256B), you might want to consider more documentation on `program` regarding the expected interactions between erasing and writing and whether or not any preservation of pre-existing content is expected.  I don't have any specific recommendation for this PR regarding "nit2", but rather to consider the behavior of the `Flash` trait).

PR #67, opened 2021-04-19
[manifest] Add a `Manager` type for managing manifest updates

mcy:
This type is analogous to a similar class of manager types in the
Microsoft C implementation, whose primary purpose is to perform
the necessary checks for activating a manifest.

Right now, we expect the caller to track the "staging" area separately,
though it may wind up making sense to move that into `Manager` in the
fullness of time.

### DIFF
--- a/src/mem/arena.rs
+++ b/src/mem/arena.rs
@@ -32,6 +32,18 @@ use crate::mem::align_to;
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct OutOfMemory;
 
+impl OutOfMemory {
+    /// Returns a static, mutable reference to an `OutOfMemory`.
+    ///
+    /// This function is the moral equivalent of `&mut []`; `OutOfMemory`
+    /// holds no data, so any non-null pointer is a suitable reference to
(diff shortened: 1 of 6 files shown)


mcy:
+jrvanwhy because there's some trait grossness here.

src/mem/arena.rs
```diff
@@ -32,6 +32,18 @@ use crate::mem::align_to;
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct OutOfMemory;
 
+impl OutOfMemory {
+    /// Returns a static, mutable reference to an `OutOfMemory`.
+    ///
+    /// This function is the moral equivalent of `&mut []`; `OutOfMemory`
+    /// holds no data, so any non-null pointer is a suitable reference to
+    /// `OutOfMemory`.
```
jrvanwhy, line 40: This comment is technically incorrect (the pointer also needs to be aligned, and the alignment isn't *guaranteed* to be 1). It can probably be shortened to "This function is similar to `&mut []`, as `OutOfMemory` holds no data.".

alistair23 (approved):

PR #70, opened 2021-04-27
server: Initial support for the AC-RoT

alistair23:
This PR focuses on initial support for the AC-RoT.

In terms of responding to requests the AC-RoT is pretty much identical to the PA-RoT responding to requests. To allow us to reuse the code this PR splits out the `process_request()` function into a trait and then has a generic `RoT` struct that is used by both AC-RoT and PA-RoT. As Rust doesn't have inheritance this ends up being the simplest way that I could think of to handle this.

This way in the future the AC-RoT and PA-RoT can have specific functions added (for example `send_request()`) and also add new traits to extend shared functionality.

### DIFF
--- a/src/server/ac_rot.rs
+++ b/src/server/ac_rot.rs
@@ -1,0 +1,171 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! A `manticore` "server" for a AC-RoT.
+//!
+//! This module provides structures for serving responses to a PA-RoT making
+//! requests to an AC-RoT.
+
+use crate::crypto::rsa;
+use crate::hardware;
+use crate::mem::Arena;
+use crate::net;
+use crate::server::options::Options;
+use crate::server::response::Respond;
+use crate::server::rot::Rot;
+use crate::server::Error;
+
+/// A AC-RoT, or "Active Component's Root of Trust", server.
+///
+/// This type implements the request -> response "business logic" of the
+/// PA-Rot <-> AC-RoT interaction.
+pub struct AcRot<'a, Identity, Reset, Rsa> {
+    rot: Rot<'a, Identity, Reset, Rsa>,
+}
+
+impl<'a, Identity, Reset, Rsa> AcRot<'a, Identity, Reset, Rsa>
+where
+    Identity: hardware::Identity,
+    Reset: hardware::Reset,
+    Rsa: rsa::Builder,
+{
+    /// Create a new `AcRot` with the given `Options`.
+    pub fn new(opts: Options<'a, Identity, Reset, Rsa>) -> Self {
+        Self {
+            rot: Rot::new(opts),
+        }
+    }
+}
+
+impl<'a, Identity, Reset, Rsa> Respond for AcRot<'a, Identity, Reset, Rsa>
+where
+    Identity: hardware::Identity,
+    Reset: hardware::Reset,
+    Rsa: rsa::Builder,
+{
+    fn process_request<'req>(
+        &mut self,
+        host_port: &mut dyn net::HostPort,
+        arena: &'req impl Arena,
+    ) -> Result<(), Error> {
+        self.rot.process_request(host_port, arena)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::crypto::ring;
+    use crate::hardware::fake;
+    use crate::hardware::Identity as _;
+    use crate::io::Cursor;
+    use crate::mem::BumpArena;
+    use crate::protocol;
+    use crate::protocol::capabilities::*;
+    use crate::protocol::device_id;
+    use crate::protocol::wire::FromWire;
+    use crate::protocol::wire::ToWire;
+    use crate::protocol::Header;
+    use core::time::Duration;
+
+    const NETWORKING: Networking = Networking {
+        max_message_size: 1024,
+        max_packet_size: 256,
+        mode: RotMode::Active,
+        roles: BusRole::TARGET,
+    };
+
+    const TIMEOUTS: Timeouts = Timeouts {
+        regular: Duration::from_millis(30),
+        crypto: Duration::from_millis(200),
+    };
+
+    const DEVICE_ID: device_id::DeviceIdentifier =
+        device_id::DeviceIdentifier {
+            vendor_id: 1,
+            device_id: 2,
+            subsys_vendor_id: 3,
+            subsys_id: 4,
+        };
+
+    fn simulate_request<'a, C: protocol::Command<'a>, A: Arena>(
+        scratch_space: &'a mut [u8],
+        arena: &'a mut A,
+        server: &mut AcRot<fake::Identity, fake::Reset, ring::rsa::Builder>,
+        request: C::Req,
+    ) -> Result<C::Resp, Error> {
+        let len = scratch_space.len();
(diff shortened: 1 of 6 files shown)


mcy:
This is super cool! It'll take some time to review, so please ping me on slack or similar if I fall behind!

src/server/ac_rot.rs
```diff
@@ -0,0 +1,167 @@
...
+
+impl<'a, Identity, Reset, Rsa> Response for AcRot<'a, Identity, Reset, Rsa>
+where
+    Identity: hardware::Identity,
+    Reset: hardware::Reset,
+    Rsa: rsa::Builder,
+{
+    fn process_request<'req>(
+        &mut self,
+        host_port: &mut dyn net::HostPort,
```
mcy, line 49: Any particular reason for not using `impl HostPort`? In general I've preferred to not force the use of a `dyn` unless I can't avoid it, and allow clients to switch to vtables if they need it for code size reasons. WDYT?

(Oh, apparently I wrote this code. Lol. Up to you whether to add yet another type param.)

src/server/options.rs
```diff
@@ -0,0 +1,29 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Options for a `manticore` "server" and "client" for a RoT.
+
+use crate::protocol::capabilities;
+use crate::protocol::device_id;
+
+/// Options struct for initialising a [`crate::server::pa_rot::PaRot`]
```
mcy, line 10: Doc nit: I'm not a fan of long names in documentation. Can you shorten them to PaRot and AcRot, and `use` these names under a `#[cfg(doc)]`?

src/server/response.rs
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! The ability for an RoT to respond to requests.
+
+use crate::mem::Arena;
+use crate::net;
+use crate::server::Error;
+
+/// The ability for an RoT to respond to requests.
+pub trait Response {
```
mcy, line 12: Mumble maybe this should be `Respond`? I worry about this getting mixed up with the response types from the protocol module.

src/server/rot.rs
```diff
@@ -0,0 +1,161 @@
...
+
+/// A RoT, or "Root of Trust", server.
+///
+/// This type implements the request -> response "business logic" of the
+/// host <-> PA-RoT interaction or the PA-RoT <-> AC-RoT. That is, it accepts
+/// input and output buffers, and from those, parses incoming requests and
+/// processes them into responses.
+///
+/// This is a generic RoT type that can be used as either a AC-RoT or a PA-RoT.
+pub(crate) struct Rot<'a, Identity, Reset, Rsa> {
```
mcy, line 30: This is probably a useful abstraction, but the PA and AC RoTs are both a little different. What's your plan for exposing this configurability concretely? (Some Rust pseudocode would be useful for illustration.)

src/server/rot.rs
```diff
@@ -0,0 +1,161 @@
+// Copyright lowRISC contributors.
+// Copyright (c) 2021 Western Digital
```
mcy, line 2: Manticore shouldn't be under WD's copyright. Sounds like a CLA issue? @asb @domrizz0

src/server/rot.rs
```diff
@@ -0,0 +1,161 @@
...
+    err_count: u16,
+}
+
+impl<'a, Identity, Reset, Rsa> Rot<'a, Identity, Reset, Rsa>
+where
+    Identity: hardware::Identity,
+    Reset: hardware::Reset,
+    Rsa: rsa::Builder,
+{
+    /// Create a new `PaRot` with the given `Options`.
```
mcy, line 42: PaRot -> Rot here.

src/server/pa_rot.rs
```diff
@@ -58,150 +34,42 @@ where
...
     use crate::io::Cursor;
     use crate::mem::BumpArena;
     use crate::net::DevicePort;
+    use crate::protocol;
     use crate::protocol::capabilities::*;
+    use crate::protocol::device_id;
     use crate::protocol::wire::FromWire;
     use crate::protocol::wire::ToWire;
     use crate::protocol::Header;
+    use core::time::Duration;
```
mcy, line 72: Looks like this wandered away from the separate block it was in? In general I've tried to keep imports from different crates in different stanzas.

src/server/rot.rs
asb: Just to clarify, per the standard lowRISC (and OpenTitan) [CLA](https://github.com/lowRISC/manticore/blob/master/CLA), contributors retain copyright. This is really about whether we maintain metadata about who contributed to a file in its header - our standard approach (used here, and in github.com/lowrisc/opentitan) is not to do so, because invariably it gets out of date and there are questions about whether a change is large enough to merit a new copyright line etc etc.

Alistair - any questions just let me or Dom know, and we can vector to the legal committee as appropriate.

mcy: (Just so I don't miss this in the future I'm setting up CI w the standard lowrisc checker, at least to flag these issues so we can sort them explicitly.)

src/server/ac_rot.rs
alistair23: Yeah, this is just based on what is already there. A trait is probably a better idea but I think that should be a separate change.

src/server/rot.rs
alistair23: This should be pretty straightforward to handle.

For things that only a PA/AC RoT can do we just add a function to `AcRot` or `PaRot` and not `Rot`. For example `PaRot` should have a `start_request()` function to initiate requests while the `AcRot` won't.

For different logic in shared functionality we can either handle it in the `AcRot`/`PaRot` specific function before calling the generic `Rot` one:

```rust
impl<'a, Identity, Reset, Rsa> Respond for PaRot<'a, Identity, Reset, Rsa>
where
    Identity: hardware::Identity,
    Reset: hardware::Reset,
    Rsa: rsa::Builder,
{
    fn process_request<'req>(
        &mut self,
        host_port: &mut dyn net::HostPort,
        arena: &'req impl Arena,
    ) -> Result<(), Error> {
        if some_rot_condition {
            return self.do_something();
        }

        self.rot.process_request(host_port, arena)
    }
}
```

Or the generic `Rot` can use the `opts.networking` to take a different action based on PA/AC RoT

src/server/rot.rs
alistair23: I'm a little confused. It sounds like it should be Copyright WD as contributors retain copyright, is that correct?

src/server/ac_rot.rs
mcy: Ack

src/server/rot.rs
asb: "Copyright lowRISC contributors" aims to catch that the work is copyrighted by contributors to this lowRISC-hosted project (intentionally slightly non-specific because moving code between repos isn't uncommon). This is not dissimilar to e.g. Go and Fuschia who use "The Go Authors" and "The Fuchsia Authors" respectively.

mcy:
Haven't had time to look at this again. Just letting you know I haven't forgotten! I'm hoping to get to it this week maybe.

alistair23:
Rebased

alistair23:
Ping!

alistair23:
Ping

alistair23:
Ping!

alistair23:
Ping!

alistair23:
Closing as this appears unmaintained

PR #73, opened 2021-04-29
[crypto] Split off signature traits to not be RSA-specific

mcy:
This will be necessary for future ECDSA support.


src/crypto/rsa.rs
```diff
@@ -4,8 +4,10 @@
 
 //! RSA, a public-key encryption algorithm.
 
+use crate::crypto::sig;
+
 #[cfg(doc)]
-use std::convert::Infallible;
+use sig::*;
```
alistair23, line 10: I know there isn't really a policy, but can we avoid using wildcard imports?

mcy: Mm, we could. I was making an exception for documentation specifically, but I think for lack of a specific policy I can explode it into a braced `use`.

cfrantz (approved):

PR #74, opened 2021-05-03
[build] Refactor the project into a Cargo workspace

mcy:
In particular, this will help with adding additional helper packages
dependent on manticore, such as an end-to-end test.

The main user-visible change is that `manticore-tool` is now invoked as `cargo r -p manticore-tool`.


Cargo.toml
```diff
@@ -58,22 +58,10 @@ std = [
   "ring/std",
 ]
 
-tool = [
-  "inject-alloc",
```
cfrantz, line 62: inject-alloc no longer appears in the tool's dependency list.  Is that intentional?

cfrantz (approved):

Cargo.toml
mcy: It does, see
```
manticore = { path = "..", features = ["inject-alloc"] }
```

This is mostly an artifact of poor choices when originally structuring the tool binary. =)

PR #75, opened 2021-05-04
Add the beginnings of proper end-to-end testing for Manticore

mcy:
`manticore-e2e` spawns a child processes to use as a "virtual RoT", speaking Cerberus over TCP. I intend to begin growing a proper conformance suite around this virtual RoT. This has three benefits:
1. e2e test for manticore itself are great.
2. We can eventually use this to test Azure's implementation as well.
3. This is intended to (eventually) be an example integration for interested Manticore users.

TODO: Add doc comments to everything.

### DIFF
--- a/e2e/src/main.rs
+++ b/e2e/src/main.rs
@@ -1,0 +1,7 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+fn main() {
+    println!("Hello, world!");
+}
--- a/e2e/src/main.rs
+++ b/e2e/src/main.rs
@@ -26,6 +26,28 @@ enum Options {
 }
 
 fn main() {
+    let pid = std::process::id();
+    env_logger::builder()
+        .format(move |buf, record| {
+            use std::io::Write;
+            for line in record.args().to_string().trim().lines() {
+                writeln!(
+                    buf,
+                    "[{level}{pid} {file}:{line}] {msg}",
+                    level = record.level().to_string().chars().next().unwrap(),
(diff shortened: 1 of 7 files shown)


e2e/src/main.rs
```diff
@@ -2,6 +2,58 @@
 // Licensed under the Apache License, Version 2.0, see LICENSE for details.
 // SPDX-License-Identifier: Apache-2.0
 
+#![deny(warnings)]
+#![deny(unused)]
+#![deny(unsafe_code)]
+//#![deny(missing_docs)]
```
alistair23, line 8: Why add this commented out?

e2e/src/pa_rot.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        Self {
+            port: 9999,
+            firmware_version: b"<version unspecified>".to_vec(),
+            unique_device_identity: b"<uid unspecified>".to_vec(),
+            resets_since_power_on: 5,
+            max_message_size: 256,
```
alistair23, line 97: Should this default be the same as the `default_value` above?

alistair23 (changes requested):
Looks good!

A few things.

 1. It's probably worth adding something to the README on how to run this, so that people know to run this as part of development.
 2. This should work well with the AC RoT work :)

PS: I have also tested Manticore with the Azure implementation for basic communication with libMCTP so it does work :)

e2e/src/main.rs
alistair23: Ah, just re-read the PR description, ignore this then.

alistair23 (approved):

cfrantz (approved):

PR #76, opened 2021-05-11
Add support for vendor_firmware_version.

osenft:
Extend the `Identity` trait with a new `vendor_firmware_version()` to support `FirmwareVersionRequest` requests with `index != 0`.

### DIFF
--- a/src/hardware/mod.rs
+++ b/src/hardware/mod.rs
@@ -20,6 +20,12 @@ pub trait Identity {
     /// Although not enforced, it is recommended that this be an ASCII string.
     fn firmware_version(&self) -> &[u8; 32];
 
+    /// Returns a string indicating the Vendor firmware version at the specified slot.
+    fn vendor_firmware_version(&self, slot: u8) -> Option<&[u8; 32]> {
+        let _ = slot;
+        None
+    }
+
     /// Returns the "unique device identity" for the device. This is a binary
     /// value of unspecified format.
     fn unique_device_identity(&self) -> &[u8];
@@ -44,24 +50,38 @@ pub trait Reset {
 pub(crate) mod fake {
     use core::convert::TryInto;
     use core::time::Duration;
+    use std::collections::HashMap;
 
     /// A fake `Identity` that returns fixed values.
     pub struct Identity {
         firmware_version: Vec<u8>,
+        vendor_firmware_versions: HashMap<u8, Vec<u8>>,
         unique_id: Vec<u8>,
     }
 
     impl Identity {
         /// Creates a new `fake::Identity`.
-        pub fn new(firmware_version: &[u8], unique_id: &[u8]) -> Self {
-            let mut firmware_version = firmware_version.to_vec();
-            while firmware_version.len() < 32 {
-                firmware_version.push(0);
+        pub fn new(
+            firmware_version: &[u8],
+            vendor_firmware_versions: &[(u8, &[u8])],
+            unique_id: &[u8],
+        ) -> Self {
+            fn pad_to_32(data: &[u8]) -> Vec<u8> {
+                let mut vec = data.to_vec();
+                while vec.len() < 32 {
+                    vec.push(0);
+                }
+                vec.truncate(32);
+
+                vec
             }
-            firmware_version.truncate(32);
 
             Self {
-                firmware_version,
+                firmware_version: pad_to_32(firmware_version),
+                vendor_firmware_versions: vendor_firmware_versions
+                    .iter()
+                    .map(|(key, value)| (*key, pad_to_32(value)))
+                    .collect(),
                 unique_id: unique_id.to_vec(),
             }
         }
@@ -70,6 +90,11 @@ pub(crate) mod fake {
     impl super::Identity for Identity {
         fn firmware_version(&self) -> &[u8; 32] {
             self.firmware_version[..32].try_into().unwrap()
+        }
+        fn vendor_firmware_version(&self, slot: u8) -> Option<&[u8; 32]> {
+            self.vendor_firmware_versions
+                .get(&slot)
+                .map(|data| data[..32].try_into().unwrap())
         }
         fn unique_device_identity(&self) -> &[u8] {
             &self.unique_id[..]
--- a/src/server/pa_rot.rs
+++ b/src/server/pa_rot.rs
@@ -76,16 +76,20 @@ where
     ) -> Result<(), Error> {
         let result = Handler::<&mut Self>::new()
             .handle::<protocol::FirmwareVersion, _>(|zelf, req| {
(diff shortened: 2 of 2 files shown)


src/hardware/mod.rs
```diff
@@ -44,33 +49,55 @@ pub trait Reset {
...
-                firmware_version.push(0);
+        pub fn new(
+            firmware_version: &[u8],
+            vendor_firmware_versions: HashMap<u8, &[u8]>,
+            unique_id: &[u8],
+        ) -> Self {
+            let mut vendor_fw_versions_vec = HashMap::new();
+            for (key, val) in vendor_firmware_versions.iter() {
+                vendor_fw_versions_vec.insert(*key, Self::u8_to_sized_vec(val));
             }
```
mcy, line 71: How about 

```rust
let vendor_firmware_versions = vendor_firmware_versions.iter().map(|(&k, v)| (k, u8_to_sized_vec(v))).collect();
```

You can even inline it into the struct constructor below!

src/hardware/mod.rs
```diff
@@ -44,33 +49,55 @@ pub trait Reset {
...
 
             Self {
-                firmware_version,
+                firmware_version: Self::u8_to_sized_vec(firmware_version),
+                vendor_firmware_versions: vendor_fw_versions_vec,
                 unique_id: unique_id.to_vec(),
             }
         }
+
+        fn u8_to_sized_vec(data: &[u8]) -> Vec<u8> {
```
mcy, line 80: I think this makes more sense as a local function in `new()`. WDYT? Also, perhaps a name like `pad_to_32` is more appropriate?

src/hardware/mod.rs
```diff
@@ -44,33 +49,55 @@ pub trait Reset {
...
+            vec
+        }
     }
 
     impl super::Identity for Identity {
         fn firmware_version(&self) -> &[u8; 32] {
             self.firmware_version[..32].try_into().unwrap()
         }
+        fn vendor_firmware_version(&self, slot: u8) -> Option<&[u8; 32]> {
+            match self.vendor_firmware_versions.get(&slot) {
```
mcy, line 96: How about:

```rust
self.vendor_firmware_versions.get(&slot).map(|data| data[..32].try_into().unwrap())
```

Although perhaps it makes more sense to make the fields of this struct use `[u8; 32]` instead of `Vec<u8>`...

src/server/pa_rot.rs
```diff
@@ -76,16 +76,21 @@ where
     ) -> Result<(), Error> {
         let result = Handler::<&mut Self>::new()
             .handle::<protocol::FirmwareVersion, _>(|zelf, req| {
-                if req.index != 0 {
-                    return Err(protocol::Error {
-                        code: protocol::ErrorCode::Unspecified,
-                        data: [0; 4],
-                    });
+                if req.index == 0 {
+                    Ok(protocol::firmware_version::FirmwareVersionResponse {
```
mcy, line 80: Given the name is used twice, maybe `use protocol::...::Response` in the closure body?

src/server/pa_rot.rs
```diff
@@ -76,16 +76,21 @@ where
...
+                    match zelf.opts.identity.vendor_firmware_version(req.index) {
+                        Some(version) => Ok(protocol::firmware_version::FirmwareVersionResponse {
+                            version,
+                        }),
+                        None => Err(protocol::Error {
+                            code: protocol::ErrorCode::Unspecified,
+                            data: [0; 4],
+                        })
+                    }
                 }
```
mcy, line 93: Maybe write this instead as

```rust
if req.index == 0 { return Ok( ... ) }
match ...
```

src/server/pa_rot.rs
```diff
@@ -257,6 +267,34 @@ mod test {
...
+        server: &mut PaRot<fake::Identity, fake::Reset, ring::rsa::Builder>,
+        request: C::Req,
+    ) -> Result<protocol::Error, Error> {
+        simulate_request_to_response::<C, A, protocol::Error>(
+            scratch_space,
+            arena,
+            server,
+            request,
+        )
+    }
```
mcy, line 296: Perhaps instead it makes sense for `simulate_request` to return `Result<Result<Resp, protocol::Error>, Error>`, which you can interrogate via the returned header's command type.

src/hardware/mod.rs
```diff
@@ -44,33 +49,55 @@ pub trait Reset {
...
 
     impl Identity {
         /// Creates a new `fake::Identity`.
-        pub fn new(firmware_version: &[u8], unique_id: &[u8]) -> Self {
-            let mut firmware_version = firmware_version.to_vec();
-            while firmware_version.len() < 32 {
-                firmware_version.push(0);
+        pub fn new(
+            firmware_version: &[u8],
+            vendor_firmware_versions: HashMap<u8, &[u8]>,
```
mcy, line 65: Maybe, in order to make calling code, easier, it makes sense to make this a `&[u8, &[u8]]`, since you're going to create a new map anyways.

src/hardware/mod.rs
osenft: Done. I actually didn't know you could have local `fn`s.

src/server/pa_rot.rs
osenft: Done.

src/server/pa_rot.rs
osenft: Done.

src/server/pa_rot.rs
osenft: Good idea, thank you! Done.

src/hardware/mod.rs
osenft: Also a good idea, thanks! Done.

mcy (approved):

src/hardware/mod.rs
```diff
@@ -20,6 +20,12 @@ pub trait Identity {
     /// Although not enforced, it is recommended that this be an ASCII string.
     fn firmware_version(&self) -> &[u8; 32];
 
+    /// Returns a string indicating the Vendor firmware version at the specified slot.
+    fn vendor_firmware_version(&self, slot: u8) -> Option<&[u8; 32]> {
```
alistair23, line 24: You can add a `#[allow(unused_variables)]` here and remove the `let _ = slot;`

alistair23 (approved):

osk:
> @osk can you squash the newest commit? Also, rebase, since that should fix CI.

Hey! Wrong mention here 😊

osenft:
> > @osk can you squash the newest commit? Also, rebase, since that should fix CI.
> 
> Hey! Wrong mention here

Haha, sorry - I usually go by "osk", but you beat me to it on GitHub.

PR #77, opened 2021-05-11
[ci] Switch to using a toolchain file to sync local and CI toolchains

mcy:
This should avoid issues with inconsistent rustfmt and clippy versions, and make a clear statement about MSRV (although Manticore reserves the right to bump the MSRV whenever is convenient).

### DIFF
--- a/src/net.rs
+++ b/src/net.rs
@@ -338,10 +338,7 @@ impl<'buf> InMemHost<'buf> {
     /// Gets the most recent response recieved until `request()` is called
     /// again.
     pub fn response(&self) -> Option<(Header, &[u8])> {
-        match self.0.tx_header {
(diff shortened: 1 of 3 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,8 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+[toolchain]
+channel = "1.52.1"
+targets = ["riscv32imc-unknown-none-elf"]
+profile = "default"
```
alistair23, line 8: Isn't this going to be out of date quickly?

PR #79, opened 2021-06-10
[cert] Add certificate parsing 

mcy:
This PR adds a generic interface for parsing certificates as part of future trust chain processing.

For now, we only have X.509 parsing, but future PRs will add OpenDice-profiled, CBOR-encoded CWT certificates. While I really, really didn't want to write an X.509 parser, it seems unavoidable as a confluence of:
- Cerberus conformance needs it.
- No existing X.509 library in Rust is pedantic in the right way; Brian Smith's `webpki` library is profiled for, well, the Web's PKI, which is a bit incompatible with what we need to do here.

~~This PR isn't done; I still need to add unit tests and fuzz tests.~~

In the name of small CLs, fuzz tests and comprehensive parser tests will be in a followup.


mcy:
Dear reviewer: this PR adds a number of test files and binary files generated from them. I suggest not reviewing anything other than the Rust files and the little script that generates the test data.

src/cert/der_test.rs
```diff
@@ -0,0 +1,333 @@
...
+        assert!(der::opt_bool(buf)?.is_none());
+        Ok(())
+    })
+}
+
+#[test]
+fn opt_bool_bad() {
+    let mut reader = Reader::new(testdata::BAD_BOOL);
+    assert!(der::opt_bool(&mut reader).is_err());
+}
```
cfrantz, line 333: Does it make sense to have tests for the `oid!` macro?

src/cert/x509.rs
```diff
@@ -0,0 +1,278 @@
...
+        der::tagged(Tag::OCTET_STRING, buf, |buf| match oid {
+            oid::KEY_USAGE => {
+                if extns.is_cert_sign.is_some() {
+                    return Err(Error::BadEncoding);
+                }
+
+                // Note that the bits in a `BIT STRING` are big endian
+                // within bytes, so we need to "shift from the left"
+                // instead.
+                const CERT_SIGN_MASK: u16 = 1 << (15 - 5);
```
cfrantz, line 237: key usage bits as defined in RFC 5280 section 4.2.1.3? (add a comment?)

cfrantz: Also, would it make sense from a namespace perspective to declare this const in the `Extensions` struct?

cfrantz (approved):

src/cert/der_test.rs
mcy: Not particularly. Its sole purpose in life is generating constants that der-ascii can emit, too.

(Also, I may wind up getting rid of `oid!` and just using der-ascii to generate the OIDs. we'll see.

src/cert/x509.rs
mcy: Done. I don't think moving it in is super necessary unless we start propagating it somehow, and even then I think it's unlikely.

PR #81, opened 2021-06-25
[fuzz] Reorganize fuzz/ to be mechanically regenerate-able

mcy:
Writing fuzz tests is a bit of a pain since it's a bit easy to trip over the generated ones. This replaces that issue with automation. This change is otherwise a no-op.

Woooo automation!

Signed-off-by: Miguel Young de la Sota <mcyoung@google.com>

PR #82, opened 2021-06-25
[cert] Add fuzz tests for the X.509 parser

mcy:
Pending on #81

### DIFF
--- a/fuzz/targets/x509_unsigned.rs
+++ b/fuzz/targets/x509_unsigned.rs
@@ -1,0 +1,54 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Fuzz tests for the X.509 certificate parser, including signature
+//! validation.
+
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+
+use manticore::cert;
+use manticore::cert::Algo;
+use manticore::cert::Cert;
+use manticore::cert::CertFormat;
+use manticore::cert::PublicKeyParams;
+use manticore::crypto::sig;
+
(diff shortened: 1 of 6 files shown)


cfrantz (approved):

fuzz/targets/x509_unsigned.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use manticore::cert;
+use manticore::cert::Algo;
+use manticore::cert::Cert;
+use manticore::cert::CertFormat;
+use manticore::cert::PublicKeyParams;
+use manticore::crypto::sig;
+
+/// A `Ciphers` that blindly accepts all signatures.
+struct NoVerify;
```
alistair23, line 20: Can you not reused the existing `NoVerify`?

fuzz/targets/x509_unsigned.rs
```diff
@@ -0,0 +1,54 @@
...
+fuzz_target!(|data: &[u8]| {
+    let _ = Cert::parse(
+        data,
+        CertFormat::RiotX509,
+        None,
+        &mut NoVerify,
+    );
+
+    // NOTE: we might actually succeed at creating a valid cert, so we can't
+    // check for is_err() here.
```
alistair23, line 53: What am I missing here? Won't this always succeed?

mcy: libfuzzer might accidentally explore a path that produces valid certificates, which is not a reason to crash the test.

fuzz/targets/x509_unsigned.rs
mcy: That would require exposing it as part of Manticore's API. If this starts happening enough I may consider putting it somewhere global, but definitely not in manticore.

PR #89, opened 2021-07-22
[protocol] Implement protocol messages necessary for authenticated measurement extraction

mcy:


src/protocol/get_digests.rs
```diff
@@ -0,0 +1,193 @@
...
+    round_trip_test! {
+        request_round_trip: {
+            bytes: &[0x01, 0x00],
+            value: GetDigestsRequest { slot: 1, key_exchange: KeyExchangeAlgo::None },
+        },
+        request_round_trip2: {
+            bytes: &[0x05, 0x01],
+            value: GetDigestsRequest { slot: 5, key_exchange: KeyExchangeAlgo::Ecdh },
+        },
+        response_round_trup: {
```
cfrantz, line 156: nit: spelling

src/protocol/get_cert.rs
```diff
@@ -0,0 +1,146 @@
...
+        let cert_number = r.read_le()?;
+        let offset = r.read_le()?;
+        let len = r.read_le()?;
+        Ok(Self {
+            slot,
+            cert_number,
+            offset,
+            len,
+        })
+    }
```
cfrantz, line 72: Is there a need for these `let` bindings?  Could you not incorporate the `r.read_le()`s into the initializer directly?

cfrantz (approved):

src/protocol/get_cert.rs
mcy: To my knowledge, Rust doesn't specify the order in which the arguments to a struct literal are evaluated. If it turns out I'm wrong, I can do a mass refactor of all the parsers, since they all do this.

PR #93, opened 2021-07-27
[mem] Allow inline arrays for BumpArena storage

mcy:
This rearranges the `unsafe` inside of `BumpArena` somewhat, but I've written comments describing why this is OK.

### DIFF
--- a/src/mem/arena.rs
+++ b/src/mem/arena.rs
@@ -7,9 +7,9 @@
 #![allow(unsafe_code)]
 
 use core::cell::Cell;
+use core::cell::UnsafeCell;
 use core::marker::PhantomData;
 use core::mem;
-use core::ptr::NonNull;
 use core::slice;
 
 use static_assertions::assert_obj_safe;
@@ -111,8 +111,7 @@ pub unsafe trait Arena {
     /// the following behavior holds:
     /// ```
     /// # use manticore::mem::*;
-    /// # let mut data = [0; 128];
-    /// # let mut arena = BumpArena::new(&mut data);
+    /// # let mut arena = BumpArena::new([0; 128]);
     /// // If the first alloc succeeds (regardless of the value of `len`), then
     /// // the subsequent alloc after resetting the arena must also succeed.
     /// assert!(arena.alloc_aligned(64, 1).is_ok());
@@ -189,14 +188,15 @@ impl<'arena, A: Arena + ?Sized> ArenaExt<'arena> for &'arena A {
     }
 }
 
-/// A bump-allocating [`Arena`] that is backed by a byte slice.
+/// A bump-allocating [`Arena`] that is backed by fixed storage.
 ///
+/// The type parameter is a type that can provide stable byte storage as long
+/// as it is not moved, such as a byte array, a byte slice, or even a [`Vec`]!
 ///
 /// # Examples
 /// ```
 /// # use manticore::mem::*;
-/// let mut data = [0; 128];
-/// let mut arena = BumpArena::new(&mut data);
+/// let mut arena = BumpArena::new([0; 128]);
 ///
 /// let buf1 = arena.alloc::<[u8; 64]>()?;
 /// let buf2 = arena.alloc_slice::<u8>(64)?;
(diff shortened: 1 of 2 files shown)


src/mem/arena.rs
```diff
@@ -337,14 +361,135 @@ unsafe impl<'arena> Arena for BumpArena<'arena> {
...
+}
+
+// SAFETY: Use the fact that Box[T] is representation-equivalent to
+// &mut T (due to FFI safety).
+#[cfg(feature = "std")]
+unsafe impl Buf for Box<[u8]> {
+    #[inline]
+    fn as_ptr(this: &UnsafeCell<Self>) -> *mut u8 {
+        unsafe {
+            Buf::as_ptr(&*(this as *const _ as *const UnsafeCell<&mut [u8]>))
```
Manishearth, line 443: this feels iffy to me; is it safe to materialize an `&UnsafeCell<&mut [u8]>` here?

src/mem/arena.rs
```diff
@@ -337,14 +361,135 @@ unsafe impl<'arena> Arena for BumpArena<'arena> {
...
+    }
+
+    #[inline]
+    fn len(this: &UnsafeCell<Self>) -> usize {
+        unsafe { (&*this.get()).len() }
+    }
+}
+
+// SAFETY: Use the fact that Box[T] is representation-equivalent to
+// &mut T (due to FFI safety).
```
Ralith, line 437: Is this meant to say `Box<T>` or `Box<[T]>`? In either case, the claim this comment makes is inaccurate:

https://doc.rust-lang.org/std/boxed/index.html#memory-layout

> **So long as T: Sized**, a Box<T> is guaranteed to be represented as a single pointer and is also ABI-compatible with C pointers (i.e. the C type T*)

(emphasis added)

Notably, `[u8]` is not `Sized`, and nor is an arbitrary `T`.

src/mem/arena.rs
```diff
@@ -337,14 +361,135 @@ unsafe impl<'arena> Arena for BumpArena<'arena> {
     }
 }
 
+/// A type that can serve as a buffer an arena can allocate from.
```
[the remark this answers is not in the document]
mcy: There's a couple of issues:
- AsRef does not guarantee that the returned slice is stable assuming the self value isn't moved.
- Holding onto a pointer into the storage doesn't work, because if the storage is inline, moving the arena will invalidate it. There's no way to detect this.

src/mem/arena.rs
mcy: Ugh. That's frustrating. I may need to just materialize a `Box` until the relevant `*mut [u8]` methods land in stable...

src/mem/arena.rs
mcy: Made irrelevant. =)

src/mem/arena.rs
mcy: Ugh. I came up with an entirely different way of handling this to avoid all this fuss.

alistair23 (approved):

src/mem/arena.rs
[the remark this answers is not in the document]
kupiakos: Wouldn't the borrow checker prevent this from occurring? How would you move a `BumpArena` while references with lifetime `'arena` are used?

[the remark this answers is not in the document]
maurer: I made a [concrete example](https://gist.github.com/maurer/14be4a82ce87293b96883a675b0bdc67) of what I meant here.

The logic as to why this works is that the `BumpArenaRef` has a lifetime restricted to that of the arena, and in turn anything allocated from that ref is restricted, so the arena should not be movable while any of these allocations are out.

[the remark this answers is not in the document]
mcy: Here's the problem, as I see it:

- I can't save the pointer to the storage in the struct at creation time, because if the struct moves, that pointer dangles.
- I can't call AsMut as necessary, because that requires an `&mut T`, which I can't create after handing out memory without violating the aliasing rules. This is why your example doesn't work, Matt.

Since neither of these solutions work, what I'm currently doing is deconstructing the storage into something I can safely get a raw pointer and a length out of, at which point I can offset things around and produce a truly unique slice reference.

jrvanwhy (approved):

PR #92, opened 2021-07-26
[e2e] Write a little test harness for e2e tests

mcy:
This will make it very easy to write lots and lots of end-to-end tests.

PR #96, opened 2021-07-28
[e2e] Make it possible to use Rust's test harness for e2e tests

mcy:
Different approach to #92, which avoids having to reinvent a test harness ourselves.

### DIFF
--- a/e2e/run.sh
+++ b/e2e/run.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Run Manticore's e2e tests. See the README.md for more details.
+
+# First, build the e2e binary itself, which acts as the server.
+#
+# --message-format=json is stable, and prints out newline-separated
+# JSON blobs that describe each cargo action. The action that outputs
+# the built binary will contain a `"executable":"<path>"` entry.
+export MANTICORE_E2E_TARGET_BINARY="$(
+  cargo build -p e2e --message-format=json \
+    | tr '\n' ' ' \
+    | perl -pe 's/^.+"executable":"(.+?)".+$/$1/g'
+)"
+
+# Now run the tests!
+# We forward the script arguments directly to the harness, to allow
+# for filtering. RUST_LOG=info can be used to enable log output.
+cargo test -p e2e -- $@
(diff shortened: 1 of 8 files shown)


e2e/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Run Manticore's e2e tests. See the README.md for more details.
+
+# First, build the e2e binary itself, which acts as the server.
+export MANTICORE_E2E_TARGET_BINARY="$(
+  cargo b -p e2e --message-format=json \
+    | tr '\n' ' ' \
+    | perl -pe 's/^.+"executable":"(.+?)".+$/$1/g'
```
cfrantz, line 9: As much as I'd love to complain about perl, the python solution here is also ugly:

```
python -c "import json,sys;print([x for x in [json.loads(line).get('executable') for line in sys.stdin] if x is not None][0])"

## or

python -c "import json,sys;print(list(filter(bool,[json.loads(line).get('executable') for line in sys.stdin]))[0])"
```

There's probably some nice solution with `jq`, but I don't really know that tool, nor is it ubiquitous.

I'm not suggesting a change, just ranting that perl will never go away.

e2e/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+# Run Manticore's e2e tests. See the README.md for more details.
+
+# First, build the e2e binary itself, which acts as the server.
+export MANTICORE_E2E_TARGET_BINARY="$(
+  cargo b -p e2e --message-format=json \
```
cfrantz, line 7: `cargo build`

e2e/run.sh
```diff
@@ -0,0 +1,13 @@
...
+    | tr '\n' ' ' \
+    | perl -pe 's/^.+"executable":"(.+?)".+$/$1/g'
+)"
+
+# Now run the tests!
+cargo t -p e2e -- $@
```
cfrantz, line 13: `cargo test`

e2e/src/pa_rot.rs
```diff
@@ -4,110 +4,63 @@
...
-    }
-
-    pub fn millis(s: &str) -> Result<Duration, ParseIntError> {
-        Ok(Duration::from_millis(s.parse()?))
-    }
-}
-
 /// Options for the PA-RoT.
-#[derive(Debug, StructOpt)]
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
```
cfrantz, line 32: Nice: -structopt +serde for communicating parameters to the child?

cfrantz (approved):

e2e/run.sh
mcy: Be happy it's not sed.

PR #99, opened 2021-08-02
[cert] Add certificate chains, which will be used in the Cerberus challenge

mcy:
~~This is a work in progress; I'll put it up for review once I add unit tests for everything and have an end-to-end test up.~~

This PR has been shortened to only add cert chains, since the previous stack had grown too tall.


mcy:
No logner set as draft, since any more PRs would make it an unreasonably large change. Tests, and the challenge itself, are in a followup (which is blocked by other changes anyways).

src/cert/chain.rs
```diff
@@ -0,0 +1,239 @@
...
+use crate::crypto::sig;
+use crate::io;
+
+/// A trust chain collection.
+///
+/// A trust chain consists of a sequence of certificates, starting with a
+/// self-signed root certificate, and ending in a leaf certificate with some
+/// a public key we wish to authenticate against the root.
+///
+/// A `TrustChain` actually managed a small number of chains across several
```
jrvanwhy, line 23: Should this be "A `TrustChain` actually manages" instead of "managed"?

src/cert/chain.rs
```diff
@@ -0,0 +1,239 @@
...
+impl<'cert, const LEN: usize> SimpleChain<'cert, LEN> {
+    /// Parses and verifies the trust chain described by `raw_chain`,
+    /// which starts at the root certificate for the trust chain.
+    pub fn parse(
+        raw_chain: &[&'cert [u8]],
+        format: CertFormat,
+        ciphers: &mut impl sig::Ciphers,
+    ) -> Result<Self, Error> {
+        if raw_chain.len() != LEN {
+            return Err(Error::Io(io::Error::BufferExhausted));
```
jrvanwhy, line 54: Returning `BufferExhausted` if a `raw_chain` that is shorter than `LEN` is passed seems misleading. Is this error code from the Cerberus specs, or is it something we can change?

src/cert/chain.rs
```diff
@@ -0,0 +1,239 @@
...
+        self.chain.get(index)
+    }
+
+    fn chain_len(&self, slot: u8) -> Option<NonZeroUsize> {
+        if slot != 0 {
+            return None;
+        }
+
+        NonZeroUsize::new(self.chain.len())
+    }
```
jrvanwhy, line 109: Nit: `cert` and `chain_len` have inconsistent style: `chain_len` has a blank line between its `if` and its normal path, while `cert` does not.

src/cert/chain.rs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+#[cfg(all(test, not(miri)))]
```
jrvanwhy, line 112: It is not at all clear to me why these are disabled when run under Miri. Is that a project-wide thing I should know? If not, please document.

testutil/src/data/keys.rs
```diff
@@ -12,10 +12,35 @@ pub const KEY1_RSA_KEYPAIR: &[u8] = include_bytes!("keys/key1.rsa.pk8");
 /// Test-only RSA public key generated from `key1.rsa.pk8`.
 #[rustfmt::skip]
 pub const KEY1_RSA_PUBLIC: &[u8] = include_bytes!("keys/key1.rsa.pub.pk8");
+/// RSA modulus of `key1.rsa.pk8`.
+#[rustfmt::skip]
+pub const KEY1_RSA_MOD: &[u8] = include_bytes!("keys/generated/key1.rsa.pub.mod");
+/// RSA exponent of `key1.rsa.pk8`.
+#[rustfmt::skip]
+pub const KEY1_RSA_EXP: &[u8] = include_bytes!("keys/generated/key1.rsa.pub.exp");
```
jrvanwhy, line 20: This suggestion is outside the scope of this PR, but maybe have a single `#![rustfmt::skip]` at the top of the file?

src/cert/chain.rs
jrvanwhy: Oh, it's because of https://github.com/lowRISC/manticore/issues/103. Perhaps leave a comment here mentioning the issue?

src/cert/chain.rs
mcy: Caught me being lazy, hehe.

testutil/src/data/keys.rs
mcy: You're probably right, honestly. I may do that in a followup. I didn't supper care bc the file is machine-generated.

jrvanwhy:
I don't see any changes since I sent my review. Do you need to push changes?

mcy:
Apparently not. Fixed now.

src/cert/chain.rs
```diff
@@ -0,0 +1,238 @@
...
+
+use core::num::NonZeroUsize;
+
+use arrayvec::ArrayVec;
+
+use crate::cert::Cert;
+use crate::cert::CertFormat;
+use crate::cert::Error;
+use crate::crypto::sig;
+use crate::io;
```
jrvanwhy, line 15: This `use crate::io;` is unused and is making CI fail.

mcy: Embarassing...

jrvanwhy (approved):

PR #109, opened 2021-08-10
[protocol] Enable zero-copy deserialization for protocol parsers

mcy:
This PR is somewhat complex, since significant plumbing needed to be done:
- First, add the `ReadZero` trait, which is analogous to the `Flash::read_direct()` function.
- Pipe `ReadZero`'s lifetime parameter through `HostPort`.
- Refactor `FromWire` to take `ReadZero`. This required doing some slightly nasty things to some tests; the tests are fine but the helpers are a bit silly. The `handler.rs` tests could use some love.

This required some substantial refactoring (most of which, thankfully, I did with a script). I can break this CL up into separate commits if desired.


src/io/read.rs
```diff
@@ -26,24 +38,89 @@ pub trait Read {
...
+impl<R: Read + ?Sized> ReadInt for R {}
+
+/// A [`Read`] that can zero-copy read data of a particular lifetime.
+///
+/// Most implementations can get away with `impl ReadZero<'_> for MyReader {}`.
+///
+/// # Safety
+///
+/// Buffers returned by `read_direct` must observe the alignment requirements
+/// requested of them. The default implementation does this automatically.
```
jrvanwhy, line 67: "the alignment requirements requested of them" is vague. How about "Buffers returned by `read_direct` must have the size and alignment specified by `layout`".

src/io/read.rs
```diff
@@ -26,24 +38,89 @@ pub trait Read {
...
+    fn read_le<I: LeInt>(&mut self) -> Result<I, io::Error> {
         I::read_from(self)
     }
 }
-assert_obj_safe!(Read);
 
-impl<R: Read + ?Sized> Read for &'_ mut R {
+impl<R: Read + ?Sized> ReadInt for R {}
+
+/// A [`Read`] that can zero-copy read data of a particular lifetime.
```
jrvanwhy, line 60: I think the wording here is misleading. They way I read it, it sounds like every instance of `ReadZero` can perform zero-copy reads, which is not the case. `ReadZero` provides an interface that supports, but does not guarantee, zero-copy reads.

The function comment on `read_direct` has the same issue. When I see `Attempts to perform a "direct read" with the given layout`, I assume that if a direct read isn't possible the read will fail, not fall back to a copying read.

jrvanwhy (approved):

PR #113, opened 2021-08-30
[protocol] Fix a build error in some configurations

mcy:
The failure is inside of a `protocol` macro, which is likely because I merged a contributor's PR in a bad way that CI failed to catch. See:
```rust
error: expected identifier
  --> src/protocol/macros.rs:89:13
   |
89 |               $field_vis $field: $field_ty,
   |               ^^^^^^^^^^
   | 
  ::: src/protocol/get_host_state.rs:63:1
   |
63 | / make_fuzz_safe! {
64 | |     /// The [`GetHostState`] response.
65 | |     #[derive(Clone, Copy, PartialEq, Eq, Debug)]
66 | | #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
...  |
69 | |     }
70 | | }
   | |_- in this macro invocation
```

GH's presubmits seem to be more of a suggestion than a hard guardrail sometimes...

Signed-off-by: Miguel Young de la Sota <mcyoung@google.com>

### DIFF
--- a/src/protocol/get_host_state.rs
+++ b/src/protocol/get_host_state.rs
@@ -23,7 +23,7 @@ use libfuzzer_sys::arbitrary::{self, Arbitrary};
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
-/// A command for requesting the host reset state
+/// A command for requesting the host reset state.
 ///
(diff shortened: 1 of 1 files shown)


jrvanwhy (approved):
Can you post the error message that this resolves, perhaps as a PR comment?

PR #115, opened 2021-08-31
[fuzz] Make make_fuzz_safe! compose better

mcy:
make_fuzz_safe! is used to generate 'static versions of structs, so
that they can derive() the Arbitrary trait. Unfortunately, the existing
mechanism was quite hacky, and did not make appropriate use of the trait
system. This commit refactors it into a trait-friendly version.


mcy:
Dear reviewer: all of the regenerated Rust files in fuzz/ can be ignored.

jrvanwhy (approved):

alistair23 (approved):

PR #121, opened 2021-09-08
[protocol] Parse manifest flags into structs

### LINKED ISSUE #60: Provide flags structs for `manifest::owned::Pfm`
Right now, they're just raw bytes. We aught to make it possible to specify them as explicitly as enumeration values or similar.

gorup:
fixes #60 

Hi! This PR seems OK to me, but likely it will just be the starting point. Obviously feedback is appreciated, and here are some thoughts I had:

- Existing code doesn't seem to use `new` often, I could reduce the usages that I introduce?
- Unsure exactly how `manifest::owned::pfm` and the elements in `manifest::pfm` really related to each other, should I make changes to `manifest::pfm` as well?
- I specifically wrote the code to be forward/backward compatible with new flags, `wire_enum` didn't seem to be a perfect fit here especially because these flags don't have the whole byte, just a bit or two at the moment.
- The `zeroth_bit_zero` and the `<< 6 >> 6` shifting don't seem great.. but they're relatively simple so I thought why try something more complicated
- I could pull these things out to a `flags` module if you'd like, the code is a bit messy all together
- Yes my tests test all `u8` values, but its not that many. I heard someone once say "why not test them all?" when referring to testing variants on numbers less than like 32 bits, so I just went along with that logic.

Thx!

### DIFF
--- a/src/manifest/mod.rs
+++ b/src/manifest/mod.rs
@@ -343,3 +343,275 @@ pub mod provenance {
         const AUTHENTICATED: bool = false;
     }
 }
+
+wire_enum! {
+    /// A policy for responding to verification failure in a read-write region.
+    ///
+    /// Cerberus currently does not fully specify what these policies mean
+    /// precisely, nor what failure mode they should be enacted with respect
+    /// to.
+    pub enum RwFailurePolicy: u8 {
+        /// Do nothing.
+        DoNothing = 0b00,
+        /// Restore the region from write-only memory.
+        RestoreFromRo = 0b01,
+        /// Erase the region.
+        Erase = 0b10,
+    }
+}
+
+impl RwFailurePolicy {
+    /// Returns the `RwFailurePolicy` flag based on a `u8`
+    pub fn from_u8(n: u8) -> RwFailurePolicy {
+        match n & 0b11 {
+            0b01 => RwFailurePolicy::RestoreFromRo,
+            0b10 => RwFailurePolicy::Erase,
+            _ => RwFailurePolicy::DoNothing,
+        }
+    }
+
+    /// Returns the input `u8` with the bits of this `RwFailurePolicy` set in it.
+    pub fn set_bits_in_u8(&self, n: u8) -> u8 {
+        (n & !0b11) | (*self as u8)
+    }
+}
+
+wire_enum! {
+    /// Indicates when this image's signature should be validates, either
+    /// only if there are changes to the image or otherwise on every boot.
+    pub enum MustValidate: u8 {
+        /// Only validate on updates
+        OnlyOnUpdates = 0b0,
+        /// Validate on every boot
+        OnEveryBoot = 0b1,
+    }
+}
+
+impl MustValidate {
+    /// Returns the `MustValidate` flag based on a `u8`
+    pub fn from_u8(n: u8) -> MustValidate {
+        if n & 0b1 == 0b0 {
+            MustValidate::OnlyOnUpdates
+        } else {
+            MustValidate::OnEveryBoot
+        }
+    }
+
+    /// Returns the input `u8` with the bits of this `MustValidate` set in it.
+    pub fn set_bits_in_u8(&self, n: u8) -> u8 {
+        (n & !0b1) | (*self as u8)
+    }
+}
+
+wire_enum! {
+    /// Indicates if the firmware component update can take affect at run-time
+    /// or if a reboot is required.
+    pub enum UpdatesTakeEffect: u8 {
+        /// Only on reboot
+        OnlyOnReboot = 0b0,
+        /// At run-time and on reboot
+        AtRuntimeAndOnReboot = 0b1,
+    }
+}
+
+impl UpdatesTakeEffect {
+    /// Returns the `UpdatesTakeEffect` flag based on a `u8`
+    pub fn from_u8(n: u8) -> UpdatesTakeEffect {
+        if n & 0b1 == 0b0 {
+            UpdatesTakeEffect::OnlyOnReboot
+        } else {
+            UpdatesTakeEffect::AtRuntimeAndOnReboot
+        }
+    }
+
+    /// Returns the input `u8` with the bits of this `UpdatesTakeEffect` set in it.
+    pub fn set_bits_in_u8(&self, n: u8) -> u8 {
+        (n & !0b1) | (*self as u8)
+    }
+}
+
+#[cfg(test)]
+#[allow(unused)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn operation_on_failure_set_bits_in_u8_do_nothing() {
+        assert_eq!(4, RwFailurePolicy::DoNothing.set_bits_in_u8(4));
+        assert_eq!(12, RwFailurePolicy::DoNothing.set_bits_in_u8(12));
+        assert_eq!(128, RwFailurePolicy::DoNothing.set_bits_in_u8(128));
+
+        for i in 0..=u8::MAX {
+            let expected = i & !0b11;
+            assert_eq!(expected, RwFailurePolicy::DoNothing.set_bits_in_u8(i));
+        }
+    }
+
+    #[test]
+    fn operation_on_failure_set_bits_in_u8_restore_from_ro() {
+        assert_eq!(1, RwFailurePolicy::RestoreFromRo.set_bits_in_u8(2));
+        assert_eq!(5, RwFailurePolicy::RestoreFromRo.set_bits_in_u8(4));
+        assert_eq!(13, RwFailurePolicy::RestoreFromRo.set_bits_in_u8(12));
+        assert_eq!(129, RwFailurePolicy::RestoreFromRo.set_bits_in_u8(128));
+
+        for i in 0..=u8::MAX {
+            let expected = (i & !0b11) + 1;
+            assert_eq!(
+                expected,
+                RwFailurePolicy::RestoreFromRo.set_bits_in_u8(i)
+            );
+        }
+    }
+
+    #[test]
+    fn operation_on_failure_set_bits_in_u8_erase() {
+        assert_eq!(2, RwFailurePolicy::Erase.set_bits_in_u8(2));
+        assert_eq!(6, RwFailurePolicy::Erase.set_bits_in_u8(4));
+        assert_eq!(14, RwFailurePolicy::Erase.set_bits_in_u8(12));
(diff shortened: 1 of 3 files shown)


src/manifest/owned/pfm.rs
```diff
@@ -102,40 +93,205 @@ pub enum Element {
...
+    DoNothing,
+    /// Restore region from RO
+    RestoreRegionFromRo,
+    /// Erase region
+    EraseRegion,
+}
+
+impl OperationOnFailure {
+    fn new(raw_flags: u8) -> Self {
+        match raw_flags << 6 >> 6 {
```
mcy, line 215: I'd have written `raw_flags & !0b11`. The shifts aren't super easy to read imo.

src/manifest/owned/pfm.rs
```diff
@@ -408,6 +564,53 @@ mod test {
         );
     }
 
+    #[test]
+    fn parse_firmware_flags() {
```
mcy, line 568: I see what you're doing here, but unfortunately it only means you test round-trips. You aren't testing that specific known-good bit patterns parse to good memory representations, and vice-versa. You'll need to add tests for those, and you can't really "test all of them" this way. =(

The round-trip-tests are, IMO, less useful.

mcy (commented):
About `manifest::pfm`...

This is spelled out here: https://github.com/lowRISC/manticore/blob/master/src/manifest/mod.rs#L64

But to be more specific, each manifest has two parsers: one that "owns" its content (and thus can be serde-deserialized) and one that is lazily parsed from a flash bank. You've implemented the former but still need to implement the latter.

I think it makes the most sense to move your flags structs into `manifest::pfm` and then refer to them from `manifest::owned::pfm`.

Also, I'm a huge stickler for documentation comments; being well-documented is a strategic goal for Manticore. I'm happy to help out offline with writing them but I think you should look at how I've written comments throughout the library and try to emulate that style.

src/manifest/owned/pfm.rs
```diff
@@ -102,40 +93,205 @@ pub enum Element {
...
+
+#[cfg(feature = "serde")]
+impl Serialize for FirmwareFlags {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        crate::serde::se_bin(self.raw_flags, serializer)
+    }
+}
```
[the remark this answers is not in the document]
gorup: - Ack on moving to `manifest::pfm` + using the flags there
- Ack on docs/comments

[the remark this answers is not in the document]
gorup: Yeah after refactoring to just hold the `u8` and using getters, this will be replaced!

src/manifest/owned/pfm.rs
```diff
@@ -102,40 +93,205 @@ pub enum Element {
     },
 }
 
+/// Firmware Flags
+#[allow(missing_docs)]
+#[derive(Clone, PartialEq, Eq, Debug, Hash)]
+pub struct FirmwareFlags {
```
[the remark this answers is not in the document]
gorup: Ack on invariant enforcing w/ getters. Thx!

One question tho is what if we are passed flags that we don't recognize? For example, they create a new flag in the 2nd, 3rd.. Nth bits, but someone is running an old version of manticore that only recognizes the 0th bit?

I tried to write my code such that we would always store the raw bits, including any we are unaware of, and be able to serialize/de-serialize all of these potentially unknown flags.

So between deserializing into only the bits we recognize, and potentially serializing only the known ones and dropping the unknown ones, versus keeping the `raw_bits`, I decided to go with the latter. But again, I could do this while keeping `raw_flags` private such that there is no potentially discrepancy between our flags, and what someone sees from `raw_flags`

gorup:
I just did pretty much an entire re-write of the logic, so a lot of the old comments don't apply.

- Addressed the invariant situation
- Tried to add better documentation comments per the info I had available, bit if you had more background you'd like me to add, let me know!
- updated bit logic
- tests have many more cases
- add getters, and setter on the `owned` PFM

Thanks!

src/manifest/mod.rs
```diff
@@ -343,3 +343,275 @@ pub mod provenance {
...
+        /// Restore the region from write-only memory.
+        RestoreFromRo = 0b01,
+        /// Erase the region.
+        Erase = 0b10,
+    }
+}
+
+impl RwFailurePolicy {
+    /// Returns the `RwFailurePolicy` flag based on a `u8`
+    pub fn from_u8(n: u8) -> RwFailurePolicy {
```
mcy, line 365: This is definitely interesting... at some point I may consider expanding the wire_enum! macro to simplify generation of bitfields like this one.

src/manifest/mod.rs
```diff
@@ -343,3 +343,275 @@ pub mod provenance {
...
+        match n & 0b11 {
+            0b01 => RwFailurePolicy::RestoreFromRo,
+            0b10 => RwFailurePolicy::Erase,
+            _ => RwFailurePolicy::DoNothing,
```
mcy, line 369: Do we want to maybe return Option here to capture that this can be out-of-bounds? I don't like silently dropping it...

src/manifest/mod.rs
```diff
@@ -343,3 +343,275 @@ pub mod provenance {
...
+        }
+    }
+
+    /// Returns the input `u8` with the bits of this `RwFailurePolicy` set in it.
+    pub fn set_bits_in_u8(&self, n: u8) -> u8 {
```
mcy, line 374: Just take `self`; it's `Copy`.

src/manifest/owned/pfm.rs
```diff
@@ -102,6 +81,49 @@ pub enum Element {
     },
 }
 
+/// asdf
```
mcy, line 84: 🤔

src/manifest/owned/pfm.rs
```diff
@@ -114,10 +136,25 @@ pub struct Rw {
...
         )
     )]
-    pub flags: u8,
+    flags: u8,
     pub region: Region,
 }
 
+impl Rw {
+    /// Gets the `RwFailurePolicy` flag value.
+    pub fn get_rw_failure_policy(&self) -> RwFailurePolicy {
```
mcy, line 145: `s/get_//`; in general, Rust getters don't begin with `get`.

src/manifest/owned/pfm.rs
```diff
@@ -102,6 +81,49 @@ pub enum Element {
...
+    )]
+    pub firmware_id: Vec<u8>,
+    #[cfg_attr(
+        feature = "serde",
+        serde(
+            deserialize_with = "crate::serde::de_radix",
+            serialize_with = "crate::serde::se_bin",
+        )
+    )]
+    flags: u8,
```
mcy, line 109: Unfortunately, this means that this can't be constructed by external crates now... not sure what the best answer here is unfortunately. We could add a `new` function that takes all the parameters... but then I'd insist on accessors for the two other fields (which we'd then make private.)

src/manifest/owned/pfm.rs
```diff
@@ -102,6 +81,49 @@ pub enum Element {
...
+    pub fn get_updates_take_effect(&self) -> UpdatesTakeEffect {
+        UpdatesTakeEffect::from_u8(self.flags)
+    }
+
+    /// Sets the `UpdatesTakeEffect` flag value.
+    pub fn set_updates_take_effect(
+        &mut self,
+        updates_take_effect: UpdatesTakeEffect,
+    ) {
+        self.flags = updates_take_effect.set_bits_in_u8(self.flags);
```
mcy, line 123: Maybe `set_bits` wants to take `&mut u8` instead?

mcy (commented):
Looking pretty good so far! Thanks for incorporating my suggestions.

src/manifest/owned/pfm.rs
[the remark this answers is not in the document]
gorup: resolving, outdated

src/manifest/owned/pfm.rs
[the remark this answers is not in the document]
gorup: resolving, outdated

src/manifest/owned/pfm.rs
gorup: resolving, outdated

src/manifest/owned/pfm.rs
gorup: resolving, outdated

src/manifest/mod.rs
gorup: yeah, I thought about that as well. The spec says "reserved (do nothing)".

Would you want me to just add that as another variant so we are fully expressing the variants?

Also, we could have a `fn do_nothing(&self) -> bool` which would be true for `0b00` and `0b11`. 

Thoughts?

src/manifest/mod.rs
gorup: yeah, we would need to allow that macro to take as input, the position of the flag in the enum.

So if there is another flag covering bits 5,6, the macro would be like:
```
// Yes the syntax is totally wrong, i'm not well read on macro syntax..
wire_bitflag! {
   position = 5..6,
   type = u8,
   enum Foo {
       bar = 0b00,
       ...
   }
}
```

src/manifest/mod.rs
gorup: done

src/manifest/owned/pfm.rs
gorup: 🤦‍♂️

src/manifest/owned/pfm.rs
gorup: done!

src/manifest/owned/pfm.rs
gorup: done

src/manifest/owned/pfm.rs
gorup: Why can't other crates construct them? 

We could have a function on the `Element`, but that's always going to yield an awkward experience. This example sets the flags, if its an `Element::AllowableFw` variant:

```
impl Element {
    fn set_update_takes_effect(&self, update_takes_effect: UpdateTakesEffect) -> bool {
    if let Element::AllowableFw  {
       version_count,
       firmware_id,
       mut ref flags,
     }  = self {
        *flags = update_takes_effect as u8;
         true
   } else {
         false
  }
}
```

But that's not great because its weird if you don't have an `AllowableFw`. Thoughts?

PR #123, opened 2021-09-13
[debug] Add manticore::debug

kesyog:
Started with PR #108 and made some non-structural changes to get it working:

* Rename `trace!` to `fail!` to avoid naming conflict with `log::trace`.
* Fix some minor syntax errors to get the macros to compile and work properly.

A couple of things I wasn't totally sure about:
1. Am I correct that the advantage of the redactable log macros (`info!`, `error!`, etc.) over the base `log` crate is that the log crate still has some overhead ("an integer load, comparison and jump") even if no logger is set up?
2. @mcy said in his original PR that the switch to the new `Result` type might not be able to be done gradually. Any particular reason why that would be?


mcy:
> Am I correct that the advantage of the redactable log macros (info!, error!, etc.) over the base log crate is that the log crate still has some overhead ("an integer load, comparison and jump") even if no logger is set up?

It's not the overhead, but rather being able to compile out the log format messages altogether; it is useful for us to be able to say "if you flip this flag, formatting is redacted at the linker level".

src/debug.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /// Transforms the wrapper error by way of an [`Into`] conversion.
+    ///
+    /// Generally, this function should not be necessary, because Manticore-
+    /// defined error types manually implement the relevant [`From`]
+    /// implementations for `manticore::Error`, which in turn call `cast()`.
+    pub fn cast<F>(self) -> Error<F>
+    where
+        F: From<E>,
```
mcy, line 47: Can we move this bound into the param declaration? (I might have put it here, in which case i appologize for the runaround =P)

src/debug.rs
```diff
@@ -0,0 +1,204 @@
...
+    {
+        Error {
+            inner: self.inner.into(),
+        }
+    }
+
+    /// Gets the wrapper error.
```
mcy, line 54: wrapper -> wrapped

src/debug.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+}
+
+impl<E> AsMut<E> for Error<E> {
+    fn as_mut(&mut self) -> &mut E {
+        &mut self.inner
+    }
+}
+
+#[cfg(feature = "log")]
```
mcy, line 72: Hmm why is this conditioned on "log"? This might be my doing but I think we can drop it.

mcy (commented):
This is awesome! Now, the question is... do you want to attempt to migrate the entire crate over to `manticore::Result`? Unfortunately I think that needs to be done in one shot. =/

We can merge this and do it in a separate CL, or you can give it a shot in this one. I suspect the way you'll wind up doing this is:
- Replace `Result` with `crate::Result` throughout.
- Add `fail`, `check` and `cast` where necessary until the crate compiles again.

If you can think of a way to do this gradually that would be even more awesome, because then you won't be rebasing over everyone else and I can help you get it done faster.

src/debug.rs
kesyog: That was my doing. My thought was to further reduce the number of compiled-in strings, though this has way less impact than the log macros. Moving the gate inside the function implementation rather than gating the entire trait implementation would probably be better. Happy to leave it out entirely if it's over-optimization at this point.

kesyog: Removed

src/debug.rs
```diff
@@ -0,0 +1,215 @@
...
+///     }
+/// }
+/// ```
+///
+/// Unfortunately, the trait coherence rules mean that for `E1 == E2`,
+/// this conflicts with the standard library's `impl<T> From<T> for T`
+/// impl. We work around this by calling this macro for every Manticore
+/// error definition that has `From` impls.
+macro_rules! debug_from {
+    ($e:ident<$trait:ident: $bound:path> => $($f:ty),+ $(,)?) => {$(
```
kesyog, line 101: This first macro rule is an ugly hack so that this macro can be used with `Error<Header>` in src/server/handler.rs. [See usage](https://github.com/lowRISC/manticore/pull/123/files?authenticity_token=G3yXZk0zkJbfp64g1zPlHUGY-1JgnaE_n28D_LkqSyzV6OfxWXs2dp4q6S7mZ--4mIwZGFZVJ4KwtjeJdfQQfg&file-filters%5B%5D=.rs#diff-4c9251cd3e4426cdcd6c86b12087130f3b5a7cf27db998d8dbb648d3de6c7237R126). It works... but I'm not too proud of it. I'm super new to Rust macros, but based on what I've read, there's not really a clean way of doing this besides resorting to proc macros, which seems like a rabbit hole not worth tackling yet. That said, feedback would be appreciated if you think otherwise.

kesyog:
Hey @mcy, while looking into using the new `manticore::Result` type everywhere, I think I found a way to switch over to the new `fail!` and `check!` macros somewhat incrementally. Check out the latest commit.

The approach is to add a few more `From` implementations to `debug_from!` so that the `?` `Try` operator still works everywhere. In order to get it to compile again, I still need to fix a bunch of instances where `Err` was being returned directly, but otherwise I think it should work.

Let me know what you think. I've moved this PR back to draft status until I fix all the compiler errors.

kesyog:
I pushed another commit that fixes most of the compiler errors. I highly recommend viewing the individual commits. Assuming this approach is acceptable, the next steps I'm envisioning are:

#### To do in this PR

* Figure out an appropriate `Result` boundary between the x509 modules and the `untrusted` crate. That crate leaks in a `core::result::Result` type and it might make sense to simply use that type to reduce the amount of error mapping.
* Fix remaining few compiler errors
* Double-check changes, especially wherever the `check!` was added, to make sure that logical errors weren't introduced.
* Run and fix unit tests if necessary

#### Future PR(s)

* Add filename and line number to the default `fail!` output
* Incrementally add error messages to the `fail!` macros
* Incrementally replace usage of `try` operator with the `fail!` macro so that error messages can be introduced.

kesyog:
After rebasing, I noticed that #148 added a new Error in protocol/spdm/error.rs that has a lifetime parameter to refer to some extra error data. I have to think a bit more on how to handle this. My initial thought is that I'd have to add a lifetime parameter to `manticore::Error` unless I threw away the extra error data or copied it.

kesyog:
Some updates:
* I added some more hackery to `debug_from` to handle error types with lifetime specifiers
* I drew a boundary at the `cert/mod.rs` level. The `parse` interface returns the new `crate::Result` type, but the underlying parsers use the old `Result` type. This reduces churn and minimizes conflict with the `untrusted` crate's return types.
* `cargo build` succeeds! 🎉 
* Tests don't compile yet, but that's next.

kesyog:
@mcy this is now ready for review and potential merge. I've held off on squashing in case it helps review, but I'll make sure to squash all the commits prefixed with "SQUASHME."

src/debug.rs
mcy: Not really. One option to cut down on repetition is to define something like

```
macro_rules! debug_from {
  (@actual $e:ident[$($tparams:tt)*] => ...) => {
    // Use `$($tparams)* as appropriate in <>
  };
  ($e:ident => ...) => {debug_from!(@actual $e[] => ...)};
}
```

Note that wherever a `<>` is allowed, you can leave it empty, e.g. `i32<>` is valid Rust. You can use this to your advantage here.

src/mem/arena.rs
```diff
@@ -204,7 +207,7 @@ impl<'arena, A: Arena + ?Sized> ArenaExt<'arena> for &'arena A {
 /// arena.reset();
 /// let buf3 = arena.alloc_slice::<u8>(64)?;
 /// assert_eq!(buf3.len(), 64);
-/// # Ok::<(), OutOfMemory>(())
+/// # Ok::<(), manticore::Error<OutOfMemory>>(())
```
mcy, line 210: Personally I actually don't think this is a very useful error: getting an error location from a failed allocation will point inside of arena code, when it would be much more useful for the caller to see the arena call that emitted the error. WDYT?

mcy (commented):
This is great work. There's a few changes I might suggest but I think those would be better left as follow-ups.

src/mem/arena.rs
kesyog: A follow-on task I'm envisioning (and that you incepted me with a few months ago 😄) is to steal ideas from https://github.com/rust-lang/rust/issues/87401: investigate whether we can use `#[track_caller]` to automatically track the locations of error conversions performed via `?`. That'd give extra, more useful debug information and would save the need to add `fail!` calls in many places.

So my thinking is:
1. Use the `Error` type with the `Arena` trait i.e. no change to current PR state.
2. If the debug information printed is not useful, callers can still manually add debug info. This is actually already done in some places in src/io/read.rs:
    ```rust
            let out = arena
                .alloc_raw(layout)
                .map_err(|_| fail!(io::Error::BufferExhausted))?;
    ```
    Since adding the fallible `Arena` trait calls are only made Since the `Arena` trait
3. Add error conversion tracking using `#[track_caller]` if possible. If it isn't, rely on step 2.

FWIW switching the `Arena` trait to using `core::result::Result` would be pretty easy and I can do that if you do feel strongly that seeing where in arena.rs the error originated would never be useful.

src/debug.rs
kesyog: I bashed my head against this a bit, but I'm struggling to get any sort of recursive calls to work as I'd expect. For now, I combined two of the macro arms by matching on the `tt` type.

kesyog:
I've squashed my commits together and this should be good to go now

src/mem/arena.rs
mcy: I don't feel strongly, if you plan on doing stuff with track_caller.

src/debug.rs
mcy: Sg. We can bash on it more later.

mcy (approved):

PR #124, opened 2021-09-22
[crypto] Replace `crypto::sha256` with the more general `crypto::hash`

### LINKED ISSUE #57: Support all types of hashes in `manticore::manifest`.
Currently, we fail if the type is anything but SHA-256; we should fix that for conformance.

mcy:
This PR is broken into four commits: one that adds the new API, two that migrate two it, and a third that deletes the old one.

Along the way, some things that previously only supported Sha256 support all three Sha strengths for free!

Fixes #57 (mostly).


src/crypto/hash.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub fn alloc(self, arena: &impl Arena) -> Result<&mut [u8], OutOfMemory> {
+        arena.alloc_slice(self.bytes())
+    }
+}
+
+/// An error returned by a hashing function.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum Error {
+    /// Indicates that the wrong size of digest was provided to
+    /// [`Engine::raw_finish()`].
```
jrvanwhy, line 56: Should this be `Engine::finish_raw()`?

src/server/pa_rot.rs
```diff
@@ -22,15 +22,15 @@ use crate::server::Error;
...
+pub struct Options<'a, Identity, Reset, Hash, Ciphers, TrustChain> {
     /// A handle to the "hardware identity" of the device.
     pub identity: &'a Identity,
     /// A handle for looking up reset-related information for the current
     /// device.
     pub reset: &'a Reset,
 
-    /// A handle to a SHA-256 engine.
-    pub sha: &'a Sha,
+    /// A handle to a hashering engine.
```
jrvanwhy, line 32: "hashing"?

jrvanwhy (approved):

PR #127, opened 2021-09-24
[server] Clean up `pa_rot.rs`

mcy:
This PR does two things:
- Moves the `PaRoT` tests to the e2e framework, which duplicates significant functionality, albeit at a higher quality.
- Factor out very large handlers into their own functions, since the indentation makes it hard to review large functions.


jrvanwhy (approved):
Much better, thank you!

PR #130, opened 2021-10-05
[net] Reorganize `manticore::net` into multiple files

mcy:
I'm about to begin implementation of AEAD support in `HostPort` for encrypted sessions; this is a preliminary cleanup to make that somewhat less painful.


src/net/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+
+//! Networking interfaces.
+//!
+//! This module provides generic, abstract networking interfaces for sending
+//! buffers over a "network". The term "network" is used very loosely: for
+//! our purposes, a network is an object upon which we can transact encoded
+//! buffers along with the bits necessary to encode a Cerberus header.
+//!
+//! For example, if an integration wished to receive requests from a "host"
+//! device over a SPI line, it tie up all the necessary implementation
```
jrvanwhy, line 13: "it tie up" -> "it should tie up" or similar

PR #134, opened 2021-10-13
[manifest] Initial parser for the PCD

mcy:
Note: this CL is layered on top of #133.

Unit tests and a `manifest::owned` implementation still pending.

### DIFF
--- a/src/manifest/pcd.rs
+++ b/src/manifest/pcd.rs
@@ -1,0 +1,610 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! The Platform Configuration Descriptor (PCD)
+//!
+//! The PCD is a signed configuration file describing what devices are
+//! connected to an ROT and how to talk to them.
+//!
+//! The [`ParsedPcd`] type is the entry-point for this module.
+
+use core::time::Duration;
+
+use zerocopy::AsBytes;
+use zerocopy::FromBytes;
+use zerocopy::LayoutVerified;
+use zerocopy::Unaligned;
+
+use crate::crypto::hash;
+use crate::manifest;
+use crate::manifest::provenance;
+use crate::manifest::provenance::Provenance;
+use crate::manifest::Container;
+use crate::manifest::Error;
+use crate::manifest::Manifest;
+use crate::manifest::ManifestType;
+use crate::manifest::Parse;
+use crate::manifest::ParsedManifest;
+use crate::manifest::TocEntry;
+use crate::manifest::ValidationTime;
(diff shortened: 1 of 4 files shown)


jrvanwhy (approved):

src/manifest/pcd.rs
```diff
@@ -0,0 +1,610 @@
...
+    pub fn rot_type(&self) -> RotType {
+        match self.data.flags & 1 {
+            0 => RotType::Platform,
+            1 => RotType::Active,
+            _ => unreachable!(),
+        }
+    }
+
+    /// Returns total number of flash ports on this RoT.
+    pub fn flash_ports(&self) -> usize {
```
moidx, line 251: what is a `flash_port`? Is it a `spi_flash` interface or something else?

src/manifest/pcd.rs
```diff
@@ -0,0 +1,610 @@
...
+        self.data.port_count as usize
+    }
+
+    /// Returns the total number of component RoTs connected to this RoT.
+    pub fn components(&self) -> usize {
```
moidx, line 256: Is a component an aggregation of endpoints connected via I2C and SPI?

src/manifest/pcd.rs
```diff
@@ -0,0 +1,610 @@
...
+}
+
+/// A SPI flash device mode.
+#[allow(missing_docs)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SpiFlashMode {
+    Dual,
+    Single,
+    DualFiltered,
+    SingleFiltered,
```
moidx, line 343: I am not sure what `DualFiltered` and `SingleFiltered` map to. I assume it is a pluton's specific hardware feature. 

There is also Quad support which may be available in some implementations.

src/manifest/pcd.rs
```diff
@@ -0,0 +1,610 @@
...
+}
+
+/// A SPI flash reset policy setting.
+///
+/// This setting describes what to do with a port's reset line when the flash
+/// is being verified with respect to a PFM.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SpiResetPolicy {
```
moidx, line 351: I find this name a bit confusing because I read it as the SPI flash reset policy, whereas in reality it is the policy of the SPI host attached to the SPI flash device. Maybe `SpiHostDeviceResetPolicy`? Anyway, I think the docstring refers to it as a `port`. I wonder if this assumes that there is MUX or something between the host and the device...

moidx (approved):
LG. I added some comments mainly for me to understand the assumptions of the implementation.

jrvanwhy (approved):

PR #136, opened 2021-10-29
[protocol] Factor out massive amounts of boilerplate into a macro

mcy:
In the past, adding new boilerplate, or new protocol structs, has been quite painful. This CL addresses it by cutting down the per-struct declarations to the bare minimum, allowing us to have one central place (a macro) for all the boilerplate to live.

### DIFF
--- a/e2e/src/tests/device_queries.rs
+++ b/e2e/src/tests/device_queries.rs
@@ -85,7 +85,7 @@ fn device_id() {
     });
 
     let arena = BumpArena::new([0; 64]);
-    let resp = virt.send_local::<DeviceId, _>(DeviceIdRequest, &arena);
+    let resp = virt.send_local::<DeviceId, _>(DeviceIdRequest {}, &arena);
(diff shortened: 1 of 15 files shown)


src/protocol/template.rs
```diff
@@ -0,0 +1,170 @@
...
+                        $rsp_to
+                    }
+                }
+            )*
+        }
+        pub use generated::*;
+    }};
+
+    (@if () { $($t:tt)* } else { $($f:tt)* }) => {$($f)*};
+    (@if ($($nonempty:tt)+) { $($t:tt)* } else { $($f:tt)* }) => {$($t)*};
```
jrvanwhy, line 169: Perhaps rename `@if` to something else, like `@internal`? The overlap with the `if` keyword confused me.

mcy: `@internal if_nonempty` captures it more completely

jrvanwhy (approved):

PR #144, opened 2021-11-05
[serde] Continue cleaning up `serde` declarations

mcy:
This PR includes:
- Use of `Borrowed` to improve serde output, by enabling the use of hex strings.
- Tests for all extant parsers.
- Aggressive use of `#[serde(with)]` modules to cut down on annotation noise by a factor of 3 or 4 in some cases.


src/serde.rs
```diff
@@ -67,59 +74,261 @@ pub fn skip_if_true(b: &bool) -> bool {
...
+#[cfg(not(feature = "std"))]
+pub fn se_hexstring<S>(
+    bytes: &impl AsRef<[u8]>,
+    s: S,
+) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let bytes = bytes.as_ref();
+    // Due to a limitation in serde, we fall back to serializing just bytes
```
jrvanwhy, line 307: As someone who isn't very familiar with Serde, I find "a limitation in serde" very vague. Could you elaborate?

src/serde.rs
```diff
@@ -67,59 +74,261 @@ pub fn skip_if_true(b: &bool) -> bool {
...
-    })?;
-    Ok(lv.into_slice())
+    #[derive(serde::Serialize)]
+    #[serde(transparent)]
+    struct HexString<'a>(#[serde(serialize_with = "se_hexstring")] &'a [u8]);
+
+    s.collect_seq(bytes.iter().map(|b| HexString(b.as_ref())))
+}
+
+/// Like `se_hexstring` but fr use with `#[serde(with)]`.
```
jrvanwhy, line 328: "fr" -> "for"

src/serde.rs
mcy: Serde requires that, when serializing a string, the entire string is buffered a priori, rather than exposing a writer-like interface like it does with sequences. Hence, I would need to allocate a buffer twice as large as `bytes` to serialize it, and I obviously can't thread an arena into this function.

src/serde.rs
mcy: Done.

jrvanwhy (approved):

PR #146, opened 2021-11-09
[protocol] Start switching to `enumflags2`

mcy:
I'm a bit concerned with how fuzzing is grafted into this; it may be prudent to upgrade our libfuzzer, or try to get a patch into arbitrary-derive that adds a way of intercepting an Arbitrary definition into one of the fields...

### DIFF
--- a/src/protocol/template.rs
+++ b/src/protocol/template.rs
@@ -13,6 +13,7 @@ macro_rules! protocol_struct {
 
         const TYPE: $CommandType:ty = $TYPE:ident;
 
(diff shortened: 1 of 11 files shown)


src/protocol/template.rs
```diff
@@ -7,12 +7,14 @@
 
 macro_rules! protocol_struct {
     (
+
```
jrvanwhy, line 10: Was this newline intentional? It doesn't look intentional.

jrvanwhy (approved):

src/protocol/template.rs
mcy: It is not.

PR #147, opened 2021-11-10
[server] Implement the start of an SPDM server

mcy:
This PR adds SPDM handlers to `PaRot`, as well as the relevant functionality to the end-to-end test framework in order to actually test it!


src/protocol/spdm/mod.rs
```diff
@@ -43,6 +46,126 @@ wire_enum! {
...
+impl Version {
+    /// The version of SPDM that Manticore implements.
+    pub const MANTICORE: Self = Self::new(1, 2);
+
+    /// Returns a new `Version` with the given parts.
+    ///
+    /// Note: `major` and `minor` should be `u4`s; the high bits are discarded.
+    pub const fn new(major: u8, minor: u8) -> Self {
+        Self {
+            version: major << 4 | (minor & 0xf),
```
jrvanwhy, line 84: Is there a test for this bit logic somewhere? Also line 149.

src/protocol/spdm/mod.rs
```diff
@@ -43,6 +46,126 @@ wire_enum! {
     }
 }
 
+/// An protocol SPDM version, consisting of a pair nybble-sized major and
+/// minor versions.
```
jrvanwhy, line 50: nit: I think there should be an "of" between "pair" and "nybble-sized".

jrvanwhy (approved):

PR #154, opened 2022-01-07
Cerberus tables

capoferro:
(depends on https://github.com/lowRISC/manticore/pull/150)

Add some simple tests for Markdown table parsing.

### DIFF
--- a/cerberus-tables/src/ast/mod.rs
+++ b/cerberus-tables/src/ast/mod.rs
@@ -11,6 +11,8 @@ use std::path::PathBuf;
 mod fmt;
 mod parser;
 
+pub use fmt::TableWithOptions;
+
 /// A Markdown file taken as parsing input.
 #[derive(Clone)]
 pub struct MarkdownFile {
--- a/cerberus-tables/src/ast/mod.rs
+++ b/cerberus-tables/src/ast/mod.rs
@@ -135,7 +135,7 @@ impl<'md> Ident<'md> {
         self.0
     }
 
-    /// Returns the textual content of the wrapped Span.
+    /// Returns the textual content of the wrapped [`Span'].
     pub fn name(self) -> &'md str {
         self.0.text()
     }
--- a/cerberus-tables/src/ast/mod.rs
+++ b/cerberus-tables/src/ast/mod.rs
@@ -1,0 +1,343 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Parsing and AST for Cerberus Tables.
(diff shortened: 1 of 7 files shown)


mcy:
Go ahead and rebase when you get a chance (I'll review as-is anyways).

mcy (commented):
I'd squash this into the previous and call it "fixed typo".

cerberus-tables/src/ast/mod.rs
```diff
@@ -271,7 +283,7 @@ pub enum TypeKind<'md> {
 
 /// A row of a [`Table`] with values of type `Type`.
 #[derive(Clone, Debug)]
-pub struct TableRow<'md, Type> {
+pub struct TableRow<'md, Type: std::fmt::Debug> {
```
mcy, line 286: Does this bound need to be here? I think it can be moved somewhere else.

cerberus-tables/src/ast/mod.rs
```diff
@@ -334,6 +346,12 @@ pub struct Error<'md> {
     pub kind: ErrorKind,
 }
 
+impl std::fmt::Display for Error<'_> {
```
mcy, line 349: Can you add a `use` for `std::fmt` and use just the `fmt` prefix?

cerberus-tables/src/ast/mod.rs
```diff
@@ -349,5 +367,123 @@ pub enum ErrorKind {
...
+            }
+        }
+        assert_eq!(row.name.name(), expected_name);
+        assert_eq!(row.desc.unwrap().text(), expected_desc);
+        assert_eq!(row.span.range.0, expected_cursors.0);
+        assert_eq!(row.span.range.1, expected_cursors.1);
+    }
+
+    #[test]
+    fn markdownfile_parse_tables() {
```
mcy, line 415: Something I might recommend for when we add more tests is to do something like rustc's UI tests: https://github.com/rust-lang/rust/blob/master/src/test/ui/fn/fn-recover-return-sign2.rs#L5

The idea being that this makes it easier to find spans instead of having to hand-code them, which can be quite painful. You can pull in files via `include_str!()`.

cerberus-tables/src/ast/mod.rs
```diff
@@ -135,7 +135,7 @@ impl<'md> Ident<'md> {
         self.0
     }
 
-    /// Returns the textual content of the wrapped Span.
+    /// Returns the textual content of the wrapped [`Span'].
```
mcy, line 138: Should be `` [`Span`] ``; note both are backticks. Rustdoc parses these automatically.

mcy:
Also: could you rename your commits to all start with `[tables]`? Thanks!

